PR #123, opened 2021-09-28
Suggested edit on PR #115

jimmychu0807:
@sacha-l 
Suggested edit on PR #115. I added a diagram where we explain the relationship between runtime and the outer node. It helps illustrate the idea we are explaining, but the style/color tone may not fit the article style.

If you confirm the diagram is good, I can open an issue to design team to redraw that diagram.


sacha-l (approved):
Thanks - good idea on the diagram!

Issue #150, opened 2021-09-30
[docs] Add XCM / XCMP article 


sacha-l:
Add an article for parachain builders. Work in progress.

NukeManDan:
Add info on using public good parachains. 
Tracking:
- [ ] https://github.com/paritytech/cumulus/issues/908

bernardoaraujor:
any estimations on when will this be available?

sacha-l:
I think the best we can do is have a page that explains basic XCM principles and how to use them, and link to all other current resources on XCM. For more advanced things, we could maintain a Q&A list - what do you think @NukeManDan @shawntabrizi ?

NukeManDan:
stub is included in #1140 , still at this time not in a place I would consider user facing

bernardoaraujor:
this [Audit Report on XCMv2](https://blog.quarkslab.com/resources/2022-02-27-xcmv2-audit/21-12-908-REP.pdf) could be a really good reference for XCM documentation.

PR #151, opened 2021-09-30
Review homepage

sacha-l:
Minor edits. Removed parachains under "Ways to Build" until we have the content from #150.


imadarai (approved):
lgtm

PR #189, opened 2021-10-01
Reorder tuts

jimmychu0807:
@sacha-l I reorder tuts in the tut index page in the order which I think a Substrate learner should follow. Though it is just a a recommended order, I still think it is better to be ordered.


sacha-l (approved):
Like it, nice touch.

PR #254, opened 2021-10-09
Review Substrate Connect

sacha-l:
Edits to the Substrate Connect article based on Pierre's feedback. Please review.


NukeManDan:
(note I changed the target branch to `main` per convention now, you can change in editing the title of the PR and a dropdown shows up)

raoulmillais:
Some more consistency suggestions:

* "light client(s)" are a technology, it is a shortening of "light blockchain client" because the client only verifies the block headers. It is  not a product or brand so should not be hyphenated or capitalised outside of usual grammatical rules.
* "Substrate Connect" is the product app developers interact with and is not hyphenated. It should be capitalised if it's converted to "English" from it's npm name. I try to always refer to it using its npm name `@substrate/connect` but that can be awkward in headings / links etc

PR #313, opened 2021-10-12
Move to const version variable in config (target develop)

nuke-web3:
This is just a tweak to make it easier to change things as needed on version updates. It is not exhaustive. Replaces #311


NukeManDan:
Sure - can add this, although I think a refactor of how we do versioning is needed... we have _many_ things with divergent versions that we should try and think about restricting docs around.

PR #368, opened 2021-10-15
Lg/proof of existence

lsgunnlsgunn:
Rough first pass at revising the content. Not super happy with it, but would like some eyes on the draft.


lsgunnlsgunn:
I didn't create a "solution" but I did walk through the tutorial successfully. Not confident about my formatting of the code and I wouldn't want to lead anyone astray because it isn't properly indented!

PR #397, opened 2021-10-20
Fixing CI build error, update yarn to v3, and swapping out some deprecated packages.

jimmychu0807:
This PR fixes on the error encountered in this action:
https://github.com/substrate-developer-hub/substrate-docs/actions/runs/1355978905

This is a recurring problem that `sharp` build occasionally failed. Some more here:

- https://github.com/substrate-developer-hub/substrate-docs/actions/runs/1355987057
- https://github.com/substrate-developer-hub/substrate-docs/actions/runs/1355972494

Based on the comment in this issue: https://github.com/lovell/sharp/issues/1882#issuecomment-534266128,
we need to fix the `minipass` to version `2.7.0`. 

In addition:

1. Update `yarn` to v3.
2. Swapping out some deprecated packages.

So webdev team, please take a look at the `package.json`.

**Once this PR is merged and pull down to your local, run `yarn install`**.

---
Also note that I realize those that make a PR from another repository (this applies to all external contributors who fork this repo, make enhancement, and PR back) won't be able to access the environment secret. So they won't be able to login to the docker repo and the link-checking CI job will fail outright.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -2,10 +2,12 @@ name: Check Links
 
 on:
   pull_request:
+    # With this setting, when we open a PR to `main` and `develop, it will run the linkchecking.
+    # This also includes every commit to an open PR to `main` and `develop`.
+    # ref:
+    #   - https://github.com/open-telemetry/opentelemetry-python/issues/1370
+    #   - https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request
     branches: [ main, develop ]
-  push:
-    # always want to including test-ci for testing gh-action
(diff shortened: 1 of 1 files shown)


jimmychu0807:
@imadarai The `private` field is added back.

Regarding doing link check for external contributors. I couldn't think of a good way right now, other than manual inspection. But we can schedule to run link checking job in mid-night to safe-guard it. But I also believe the task of adding new links will mostly be done among docs team members rather than external contrib.

Other repos use another methods to perform link checking. We build a docker image and push to a repo, finally serve it, and have another process running link checker to scan. Other repo just do a static scan on links in markdown documents.

PR #403, opened 2021-10-20
[docs] Reformat lockable currency 

Eve-Parity:
To be more inline with our style guide.


Eve-Parity:
> Made a few commit suggestions to ensure the "one-sentence-per-line" rule. Also added indentation for bullets. Do we have a convention around that?

Bullets should auto-indent unless they're in a step. In that case, you have to tab once. Same with code blocks in steps.

sacha-l (approved):
@Eve-Parity : Made some changes to better organize the steps and follow our style guide for how-to guides. This primarily meant including a "Before you begin" section.

PR #422, opened 2021-10-23
Fix Ubuntu installation instructions

ptqa:
* Add `g++` package to Ubuntu installation instructions. Otherwise, cargo-build on package `librocksdb-sys` fails with `error occurred: Failed to find tool. Is c++ installed?`
* Add `zlib1g-dev` package to Ubuntu installation instructions. Otherwise, cargo-build fails on linking with 
```
  = note: /usr/bin/ld: cannot find -lz           
          collect2: error: ld returned 1 exit status
```

### DIFF
--- a/v3/docs/01-getting-started/c-installation/index.mdx
+++ b/v3/docs/01-getting-started/c-installation/index.mdx
@@ -27,7 +27,7 @@ Use a terminal shell to execute the following commands:
 ```bash
 sudo apt update
 # May prompt for location information
-sudo apt install -y git clang curl libssl-dev llvm libudev-dev
+sudo apt install -y git clang curl libssl-dev llvm libudev-dev g++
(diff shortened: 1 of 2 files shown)


sacha-l:
@ptqa @NukeManDan @shawntabrizi  not sure how to proceed here? I personally can't verify what's brought up here. Can we decide whether we want to rectify or drop this PR?

ptqa:
@sacha-l there is way to verify it - run a new ubuntu VM and check a list of packages installed by default. If required packages have no g++ & zlib then we should probably add it to docs and other scripts.

Luckily I've just did it for clean Ubuntu server  image on AWS - here is a list of packages installed by default https://termbin.com/b1w9 (output of `dpkg -l` command). And there is no g++ or zlib-dev.

PR #432, opened 2021-10-25
[docs] Simple Crowdfund rework

Eve-Parity:
This is going to need a bit of work. There is a pretty large leap between the knowledge expectations of the Randomness  How-To and this one. I'm going to need some assistance figuring out which files are being edited with the code snippets. 
@sacha-l


NukeManDan (approved):
At very long last, this was rebased and updated to ensure formatting is correct (`yarn fmt` used)

PR #448, opened 2021-10-29
Tx pool style and copy updates

NukeManDan:
tweaks to style, very minor copy edit.
uses "1 idea, one line" convention over the 100 char limit.


sacha-l:
I think I can understand where you're coming from but this would be a good opportunity to align on putting our style guide to practice. This "h3" convention doesn't match up the rest of the guides so I would lean towards consistency with the others. Thoughts on this one @lsgunnlsgunn 😃 ? Perhaps we can improve how these guidelines are articulated [here](https://docs.substrate.io/v3/contribute/templates/)..

lsgunnlsgunn:
Having a h3 follow h1 breaks a fundamental rule. Please don't do this. If you think you have a legitimate rationale for breaking this rule, please explain what it is. (If you just don't like the size of h2 or something, let's put that in the list of visual elements we might want to revisit. In most cases, most of your headings **should be** h2 level heads.

sacha-l (approved):

PR #454, opened 2021-10-29
Updated version and keywords types accross for consistency

imadarai:
Quick scan to make sure only version and keywords are updated across. 

This was done dynamically, so I just want another pair of eyes.  I've already check for `yarn dev` errors and they are gone.

### DIFF
--- a/v3/docs/00-style-and-contributor-guidelines/a-contributor-guidelines/index.mdx
+++ b/v3/docs/00-style-and-contributor-guidelines/a-contributor-guidelines/index.mdx
@@ -1,7 +1,7 @@
 ---
 title: Welcome to the Substrate community
 slug: /v3/contribute/style-guide
-version: 3.0
+version: '3.0'
 section: docs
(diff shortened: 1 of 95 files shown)


sacha-l:
LGTM - not sure what the reason to do this is though, could you clarify please ? The main change I see is adding single quotes around the version string and formatting the keywords. Do we need to document this in our contributing section also?

imadarai:
> LGTM - not sure what the reason to do this is though, could you clarify please ? The main change I see is adding single quotes around the version string and formatting the keywords. Do we need to document this in our contributing section also?

No, we don't need to document.  While were were going about making changes, we messed up the underlying types of our frontmatter.  We started using 'polkador-0.9.xxx' (a string) while it was only meant to number.  Same for keywords.  In some places we used string, while in others we used array of strings.  

I'm just normalizing and making them both strings (based on recent discussions), so the gatsby-CI stops complaining on develop and build.

PR #456, opened 2021-10-30
[main v3] glossary revision

lsgunnlsgunn:
I complete the editing pass but some changes might be more controversial than others so would like some eyes on it.
Please add suggestions for additional terms to this PR (if you think of any.

Also, this is pointed to merge to main instead of develop. I'm not sure where we landed on merging directly to main in this repo.


v3/docs/01-getting-started/e-glossary/index.mdx
```diff
@@ -7,640 +7,485 @@ category: getting started
...
 
-## Aggregation
+## aggregation
 
-Used in the context of [FRAME](#frame), "[pallet](#pallet) aggregation" means combining analogous
-types from multiple runtime modules into a single type. This allows each module's analogous types to
-be represented. Currently there are six such data types:
+Used in the context of [FRAME](#frame), _aggregation_ or _[pallet](#pallet) aggregation_ is the process of combining analogous types from multiple runtime modules into a single type. 
+Pallet aggregation allows each module's analogous types to be represented. 
+Currently, there are six such data types:
```
shawntabrizi, line 37: as a note, we also call this the "outer" object.

For example the "outer call" represents the aggregated call which is mentioned in this description. Maybe this is just nomenclature, but maybe also worth including.

shawntabrizi (approved):
looks strictly better than before

Issue #280, opened 2021-10-11
Update ToC


imadarai:
Highlights need a bit tweaking + new UI given some of the H3s are super extended. 
Need to re-write the way headings are generated and perhaps use a React-Hook instead of relying on graphql and MDX to provide the data.  
Account for undefined values

PR #458, opened 2021-10-31
[main v3] Revise the published version of the private network tutorial (pre-split)

lsgunnlsgunn:
I started down this path to simplify and streamline this one tutorial, hit a wall, but think I've found the way forward. In the process though, I started to see this tutorial as overloaded and want to split it into two parts (Simulate a network and Start a private network). I created prototypes for the split tutorials, but there might be value in having some eyes on this pre-split version if people have cycles available (the last part doesn't quite work as written and at least according to comments in the file, didn't work previously either--I think I know how to fix it, but need to test if my planned fix actually works). I'll add the split-into-two to the PoC for the doc architecture that I'm planning to build next week.


lsgunnlsgunn:
Ready to have :eyes: with a few notes:
- I want to split this into two tutorials (simulate, separate from start a private network). I did a test run in the sandbox and think it is the right way to go.
- I got everything except finalization with two nodes to work. Maybe there's a step missing? 
- I'm thinking I might remove the polkadot-js "Attach a front-end" section. It doesn't add much and maybe should be a separate tutorial about connecting to a local node/connecting to a specific network.

Eve-Parity:
> Ready to have 👀 with a few notes:
> 
> * I want to split this into two tutorials (simulate, separate from start a private network). I did a test run in the sandbox and think it is the right way to go.
> * I got everything except finalization with two nodes to work. Maybe there's a step missing?
> * I'm thinking I might remove the polkadot-js "Attach a front-end" section. It doesn't add much and maybe should be a separate tutorial about connecting to a local node/connecting to a specific network.

I definitely agree that the tutorial needs to be split up, and instinct tells me that you're correct in how you want to split it. For me the split came right at **Generate your own keys**, which makes be think we're on the same page. 
Agree with nixing the front-end part.

lisa-parity:
@sacha-l  @shawntabrizi , it would be great to have a technical review on the content (especially since I could not get finalization to work for the non-predefined accounts). I'm using the revision as the basis for splitting the tutorial into 3 parts:
1) Using predefined accounts
2) Using generated keys and a chain spec
3) Using Polkadot-JS to connect to a network or local endpoint

lsgunnlsgunn:
I'm splitting this into three parts in the sandbox-removing the UI stuff from the simulated network and private network to be a standalone topic.

lisa-parity:
I'm going to merge this as "better than the current vesion" and improve in the split up version that I have in the sandbox.

decentration:
Hi! I'd like to make some general comments about "Start a private network" tutorial, and sorry if this not the best place, Dan Shields linked me here regarding my interest in making a how-to. 

Firstly, the "Start a Private Network" has been super helpful and i still refer to it after almost a year of learning Substrate. 

If I could share some comments on my achievement milestones learning Substrate, which may perhaps be helpful. 

## Natural path of learning 
For me, general achievement milestones for the DevOps side of things went something like:

1. **Build a local chain produce blocks** _"...hurrah!"_
1. **Build a dev chain and connect nodes not just local** _"...empowering!"_
1. **Build a custom chain (runtime, pallets etc)** _"...let's dance"_
    - run a chain with Docker
    - build Docker image and customize Dockerfile _"praise be"_
1. **Build a parachain and connect it to a relay chain** _"...lets do dis!"_
    - local 
    - dev
    - produce blocks on chachacha _(current task)_ _"why won't collator blocks be produced, why why why"_
1.  **Build a rococo staging relay testnet chain**  and get validators to work. 
    - customise roococo staging config in chain_spec file to have 3 validators.
    - register keys.    🎶 _"21 session keys hanging on the wall... and if one validator key should accidentally fall..."_ 🥴
    - understand about validators, aura, babe, beefy... _(current task)_ [#280](https://github.com/paritytech/subport/issues/280) 
    


 
### How to build a public network instead of private  
A very general comment would be to drop the "private" as a main tutorial because only a small minority of beginners would be interested in building a private network, and change title to something like "How to build a test network".



### Educational workflows 
An interesting thing could be to split various workflows into beginner, intermediate, advanced. So "How to build a private network" would be super high level for _beginners_, and for _advanced_ we're learning why a lot of things are done a certain way. Easy for me to say, but would awesome.



### How to build a custom relay testnet
And secondly i would like to help create one of these workflow tutorials. I was thinking the "How to build a custom relay chain testnet" would be befitting on what im working on currently. This has pushed me to learn a bit more about the validator section of things, which i request some clarity here [#280](https://github.com/paritytech/subport/issues/280) and if you could help answer any of those questions, I would be grateful.

lisa-parity:
@decentration , thanks for the excellent feedback and suggestions. Very helpful input.

PR #465, opened 2021-11-01
Missed line

manylov:
It seems like the following line is missed in examples files
```use scale_info::TypeInfo;```
it produces compilation errors.


sacha-l (approved):
Thanks @manylov - this came recently, thank you for flagging and fixing this. 🙏🏻

PR #475, opened 2021-11-03
Update macro to use by default for pallets

NukeManDan:
> Just FYI for all pallet developers and reviewers, all new pallets should have:
> 
> ```rust
> #[pallet::pallet]
> #[pallet::generate_store(pub(super) trait Store)]
> #[pallet::generate_storage_info]            <--- THIS LINE
> pub struct Pallet<T, I = ()>(PhantomData<(T, I)>);
> 
> Which will help ensure all of your storage items are bounded
> Old pallets do not have this, and we need to migrate them, but new pallets have no excuse :)
> We should consider flipping the behavior so that it is implicit it rather than needing to include it

From @shawntabrizi :point_up:
I will also update the template & example pallet in substrate & update the templates we own in this org, once this PR is apprved.


v3/docs/03-runtime/b-macros/index.mdx
```diff
@@ -180,13 +180,39 @@ This is an example of its definition:
...
 #[pallet::pallet]
 #[pallet::generate_store(pub(super) trait Store)]
-pub struct Pallet<T>(PhantomData<T>);
+#[pallet::generate_storage_info]
+pub struct Pallet<T, I = ()>(PhantomData<(T, I)>);
 ```
 
 **Docs**
 
 - See the [macro expansion](/rustdocs/latest/frame_support/attr.pallet.html#macro-expansion-1) added to the `struct Pallet<T>` definition
```
NukeManDan, line 190: Below is new content to review

sacha-l:
@NukeManDan : AFAIK the `pub struct Pallet<T, I = ()>(PhantomData<(T, I)>);` syntax is to handle pallet instantiability - not sure why we'd want that in all tutorials ? I think what's needed is just to add the `#[pallet::generate_storage_info]` line only.

NukeManDan:
Ah - makes sense to me. Should this instantiable syntax be highlighted somewhere in the docs I am not seeing? I can update to no PhantomData.

sacha-l (changes requested):
I think instantiable pallets fall outside the scope of what this PR is trying to achieve, which is adding `#[pallet::generate_storage_info]`. On a separate note, would be nice to update [this guide](https://docs.substrate.io/how-to-guides/v3/basics/instantiable-pallets/) at some point!

sacha-l (approved):

PR #482, opened 2021-11-03
 add .prettierignore, format .mdx files, run fmt:docs #210 

NukeManDan:
2nd try of #210 -- this time no line length linting. We should be mindful in #342 to use wha the fully realized style guide ends up landing on for this. Otherwise, I think this is an easy win to add. 

To run the formatter over all the _docs files_ (`*.md` and `*.mdx`) run:
```bash
yarn run fmt:docs
```
@imadarai I left a command for formatting the code too, but have not run here. It's likely something you need to fix before using. 

### A note for reviewing:
A lot of files changed, but _no_ content changes happen in this PR. 
I have run this locally, and AFAICT it LGTM for the build. (CI too!)

PR #488, opened 2021-11-04
[docs] Pallet integration formatting fix (minor)

Eve-Parity:
There was a formatting problem with the Note.


NukeManDan:
Was this done normally or with the #482 tooling?  I would suggest we move to this to make tweaks like this automated :pray:

Eve-Parity:
The note wasn't indented in the step - not sure a linter would catch that.

sacha-l (approved):

PR #514, opened 2021-11-09
Update get started, refactor images, new reference section, tweaks to docs homepage

NukeManDan:
This was my flight home's project and got a little ambitious on the branch. :sweat_smile: 

### Included

- Moves all docs/htg/tutorial images out of static into the proper (rendered) src/img folder
- New reference section where the glossary and troubleshooting are migrated to, this will get a few more pages (like a FAQ) as well. 
- Minor changes to the style for the main landing page. 
- Migrate "installation" to" Developer Environment" as it's more fitting for what you are doing. 

### Questions

- I did _not_ update any of the non-english translations. What action is needed for this to generate them?


src/components/DevNavMenu.tsx
```diff
@@ -42,17 +42,13 @@ const DevNavMenu = {
...
-          {
-            title: `${intl.formatMessage({ id: 'docs-menu-glossary' })}`,
-            link: '/v3/getting-started/glossary/',
+            title: `${intl.formatMessage({ id: 'docs-menu-dev-env' })}`,
+            link: '/v3/getting-started/dev-env/',
           },
+          // {
+          //   title: `${intl.formatMessage({ id: 'docs-menu-windows' })}`,
+          //   link: '/v3/getting-started/windows-users/',
+          // },
```
NukeManDan, line 51: intentionally hiding windows specific page from nav.

src/components/DevNavMenu.tsx
```diff
@@ -42,17 +42,13 @@ const DevNavMenu = {
...
-            title: `${intl.formatMessage({ id: 'docs-menu-windows' })}`,
-            link: '/v3/getting-started/windows-users/',
-          },
-          {
-            title: `${intl.formatMessage({ id: 'docs-menu-glossary' })}`,
-            link: '/v3/getting-started/glossary/',
+            title: `${intl.formatMessage({ id: 'docs-menu-dev-env' })}`,
+            link: '/v3/getting-started/dev-env/',
           },
+          // {
```
NukeManDan, line 48:
```suggestion
          // Hiding windows from navbar, still rendered and pointed to in dev env page
          // {
```

sacha-l:
Nice. I think these types of changes will line up with what @lsgunnlsgunn has in mind with restructuring things. These changes look good to me. My only thought is shouldn't we synchronize first so we avoid potentially duplicating reorganization / clean-up work like this?

sacha-l:
@NukeManDan : given you've moved the "Glossary" section, you must make sure to replace all `/v3/getting-started/glossary/*` links with the new slug for the "Glossary" page. This is causing our link checker to fail as far I can see. 🎯

v3/docs/01-getting-started/c-dev-env/index.mdx
```diff
@@ -1,10 +1,10 @@
 ---
-title: Installation
-slug: /v3/getting-started/installation
+title: Development Environment
+slug: /v3/getting-started/dev-env
```
sacha-l, line 3: @NukeManDan : this type of change needs to be propagated to all other slugs that are intended to point to the same content. 

Additionally I'm unsure about using an "abbrv." in the slug. I would go for something more human readable, like "environment". @Eve-Parity @lsgunnlsgunn thoughts?

sacha-l (changes requested):
I think this PR needs to fix the broken links before we can properly review. Also, a quick review from @lsgunnlsgunn would go a long way IMO in light of synchronizing how we want to restructure things going forward. 😃

v3/docs/01-getting-started/c-dev-env/index.mdx
NukeManDan: Fixed, :crossed_fingers: the CI thinks so too. I think `dev` and `env` are effectively industry standard and it helps shorten URLs. Open to find&replace to `developer-environment`

NukeManDan:
This might be a good reference, but we are going to refactor this based on what @lsgunnlsgunn  has planned :partying_face:

PR #518, opened 2021-11-10
fmt update

NukeManDan:
3 spaces needed to indent continuing lists for prerrier to like thinks, 2 it thinks isn't proper.
 https://google.github.io/styleguide/docguide/style.html#nested-list-spacing


NukeManDan:
superseded by #528

PR #521, opened 2021-11-11
Fix numbering add contracts guide

sacha-l:
This PR reformats a few guides based on the: 

-1. ...
-1. ...
-1. ...

numbering convention. 

Worth mentioning that the `yarn fmt:docs` is breaking the numbering in the guides and tutorials that have this convention. It seems to break the "indent blocks between numbers" rule. This may be related to #528.


imadarai (approved):
Seems like just a simple tab updates.  lgtm!

PR #528, opened 2021-11-12
Fix formatting, *confirm* prettier to work, hack to fix link checker

NukeManDan:
- [x] Fixed all rendering issues #482 made\
- [x] Updated deps for formatter 
- [x] Updated all continued lists to 3 spaces (prettier enforced)
- [x] Ran `yarn fmt:docs` - no changes = all md & mdx files correctly formatted (as we have configured now) :tada: 
- [x] Ignore all URLs that included `undefined` in lick checking (locally confirmed that all links pass! :tada: )

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "main": "index.js",
   "author": "Parity Technologies WebDev Team",
+  "private": true,
   "license": "MIT",
   "scripts": {
     "clean": "gatsby clean",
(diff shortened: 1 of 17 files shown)


lisa-parity (approved):
Cool. Can you write up a simple how-to for when/where to run the formattter?

NukeManDan:
yes, it can be done by CI on all PRs, and I suggest doing it manually before doing any PR here 
`yarn fmt:docs` fixes all MD and MDX files. If you use prettier directly via extension or installed globally, you can run on any single page instead of them all.

imadarai:
@NukeManDan I've made a commit to downgrade packages so they don't conflict.  Feel free to update anything in `devDependencies`, but try to avoid production `dependencies`. 

PS it's the same reason the link-check failed.  Docker failed to launch because of yarn-lock conflict.

PR #532, opened 2021-11-13
Update index.md

juuuuuu:
- Remove useless white spaces.
- Fix the numbered list numbers.


NukeManDan:
Thanks for this :grin: but MD is rendering the numbering correctly, and it's in our style guide to do it this way, as well as we use a formatter to get things polished right, see #528

PR #542, opened 2021-11-15
Lg/sandbox

lisa-parity:
I built the sandbox I shared today locally using vanilla markdown and mkdocs as a static site generator without any fancy Galsby stuff. This is a start at moving it into substrate-docs. The community section seemed like the least controversial to experiment with (I think the site builder part might need to be updated and possibly moved outside of this structure entirely into a team process page on Notion or something like that).

The strawman top-level folders I proposed:
Welcome to Substrate (single landing page but maybe this is covered by an existing page--- I needed this to build the mkdocs mockup)
Quick start (single page for a minimal entry)
Docs (folder with subfolders with potentially multiple topics each)
Tutorials (folder with multiple topics)
Reference (folder with subfolders/links)
Community (folder with subfolders, one topic each)
FAQ dropbox (ideally just markdown files contributed by other folks)
Playground (link)

Side note that I would be in favor of simplifying around a solution like mkdocs, but realize that might make it too hard to tie in with other parts of the site.


NukeManDan (changes requested):
Content looks great! I just want to tweak how to go about working on this PR a bit:

I would propose that in the top level working dir where V3 lives, we add all the new content in `pages` to replace docs-sandbox as you have it.

And this is only in transition in a **new branch** called `sandbox` that we hookup auto deployment to apart from main and develop . Then we can submit PRs against this common target.

Before moving to main we should delete the V3 folder in sandbox, and have everything build off pages so it's a clean replacement once we are ready to make the switch (and helps the git history be clean and small too) - with a squash merge of main <-sandbox

sacha-l (approved):
I say we merge this into `sandbox` and keep using `sandbox` as our working branch, given that this was the first PR for this revamp initiative.  Going forward all PRs will be to merge into `sandbox` to avoid any mix ups and merge conflict. @NukeManDan thoughts?

PR #545, opened 2021-11-16
Anchor not working, misses trailing "s"

tk-1io:
Currently does not jump to wanted section because the anchor hash misses a traling `s`


sacha-l (approved):
Thanks for the PR @tk-1io 🙏🏻 . Took the opportunity to fix other places where this was the case. 😃

PR #562, opened 2021-11-17
[sandbox] Split private network into two tutorials

lisa-parity:
I split the private network tutorial into two tutorials 
1) using the predefined accounts
2) generating keys and a chain spec

This PR covers the first one. 
I also felt the Polkadot-JS section was a distraction. I'll put that in a third tutorial.


sacha-l (approved):
LGTM

PR #575, opened 2021-11-18
Test branch for building mkdocs friendly site out of the existing san…

lisa-parity:
…dbox branch

I was originally thinking that we wouldn't use mkdocs except for a mockup, but if we want to use it as part of this revamp for structuring purposes, maybe we don't need a separate repo, but can just add the relevant stuff to the sandbox branch. Regardless of whether we use a separate repo or just a branch, if `mkdocs` is part of the equation, it doesn't like the `mdx` files.

I did a little test in this branch--it's pretty equivalent either way--working in separate mkdocs-friendly repo or adding the mkdos stuff to our current repo in the sandbox branch. I can go wither way, I just wanted to test this out. We can close the PR if y'all think the separate repo is easier (less cluttered).

If you install mkdocs locally, you can run `mkdocs serve` for hot page reloading ... and `mkdocs build` to build locally.


NukeManDan:
Migrating work to https://github.com/substrate-developer-hub/docs-sandbox/

PR #585, opened 2021-11-21
Updated Kitties tutorial to use monthly-2021-11-1

Petras98:
Updated the Kitties documentation page to use monthly-2021-11-1.


NukeManDan (approved):
better to use this, yes.

Issue #478, opened 2021-11-03
[Content] Make compilable docs


gnunicorn:
### Content request

We want to make sure that all code examples and tutorials in the docs can be (automatically) compiled, so we can easily test that they still build. Little is worth than having people copy paste a tutorial or example and find the code not building at all because we broke that without noticing. 

Refs  

### Are you willing to help with this request?

Maybe (please specify)

NukeManDan:
I have a plan in mind for this I will outline shorty, but the gist is:
1. All are based on the node template
2. Generate changes, keep important commits in mind
3. generate patch files for each step of note to review
4. Display the contents of the _patch files_ in the docs, not hard coded `rust` blocks
5. CI will be used (manually) to apply patches, and confirm the node builds&runs
6. On upgrades if changes to patch files are needed to fix CI, we tweak as needed.

lsgunnlsgunn:
I'm all for doing includes for code inside of codes, but this seems a bit heavyweight process-wise )at least to me).

NukeManDan:
I will get a PoC made for the add nicks tutorials to demonstrate the flow here :soon:

NukeManDan:
### https://github.com/paritytech/substrate-template-generator
This is a very rough README and NOTES.md describing the the tool

Basically if a full substrate node exists, you can use this tool to "fork" it into a new independent repo in a one liner setting the config you want: https://github.com/paritytech/substrate-template-generator/tree/main/config

With this template generated using  any upstream commit, we can then with scripts/CI apply git patches for the devhub examples on top of and test them. 
So it is the foundation for the outlined workflow (above)

Issue #477, opened 2021-11-03
[Content] Pin docs to released versions


gnunicorn:
### Content request

For the time being we will be considering the substrate in the latest `polkadot-x.x.x`-release branch the most stable. Docs should thus use these as their baseline and work with these.

### Are you willing to help with this request?

Maybe (please specify)

gnunicorn:
Additionally, if possible it'd be great if users could switch to `master` as well as older previous-released versions easily – both stating that they are looking at an unsupported versions.

bkchr:
Didn't we actually decided to only have one doc page and that always points to master? To remove this entire versioning at all?

NukeManDan:
Having an _unstable_ document site on something like a daily basis to learn from is highly problematic. While not _many_ things from commit to commit, end users learning from these docs will have to update manually just to be able to continue. I am generally opposed to such a fluid environment. 

The monthly cadence is presently what is used, and that is already a rather large time-suck for the docs&support team to maintain along with the _many other versionings that need to be accounted for_:
- Substrate itself
- Polkadot & Cumulus (& [Canvas soon](https://github.com/paritytech/canvas/issues/97) )
- Frontier
- .... more "Substrate Library Extensions"
- (Not directly here, but for the community) ORML and the like 3rd party pallets

These by definition do not move together, and cause us to need different versioning anyhow. 

With #478 completed, it will be _much_ easier to maintain docs moving forward, before even considering moving to anything faster this should be implemented. Still the issue of user experience needs to be considered here regardless.

NukeManDan:
I am leaning towards Polkadot release cadence (for now) - https://github.com/substrate-developer-hub/substrate-node-template/issues/260

PR #591, opened 2021-11-21
Add Troubleshooting and FAQ section 

### LINKED ISSUE #583: Polkadot.js app does no longer have developer page in permissioned network tutorial
## Which page you are referring to?
https://docs.substrate.io/tutorials/v3/permissioned-network

## What is the problem?
1. In the [custom type mapping](https://docs.substrate.io/tutorials/v3/permissioned-network/#alice-and-bob-start-the-network) part, it seems that the `PeerId` configuration step can be skipped without affecting the rest of the tutorials.
2. Polkadot.js app is no longer has [developer page](https://polkadot.js.org/apps/?rpc=ws%3A%2F%2F127.0.0.1%3A9944#/settings/developer) in the latest version.

## What could be done?
Delete that part or give a further explanation (why we don't need that now).

NukeManDan:
This PR was originally called "Use Apps UI use without `scale-info`" .  We agreed to alter the PR and create a section for the information it was originally putting forward.

Originally it: 
- [x] Closes #583 
- [x] Adds instructions that could be a stand-alone HTG on Apps UI use without `scale-info` (in followup PR) 

With @sacha-l 's recent updates, it now:
- Adds a section in our docs called "Troubleshooting and FAQ" 
- Add a page that contains information on: how to use PolkadotJS apps with a Substrate runtime that doesn't have the new `scale-info` pallet and how to use `--dev` without the (now) default `--tmp`.

This directly addresses #631 . Going forward, this section can be used by the support team and slowly will morph into the "FAQ dump" page idea the docs team has been pondering about.

### DIFF
--- a/v3/tutorials/03-permissioned-network/index.mdx
+++ b/v3/tutorials/03-permissioned-network/index.mdx
@@ -599,7 +599,7 @@ permissioned network. You can also play with other dispatchable calls like
 > This powerful and very useful feature was introduced [in this PR](https://github.com/paritytech/substrate/pull/8615).
 > We highly recommend including this in your node!
 
-The hosted apps UI is _not compatible_ with custom metadata any longer.
+The hosted apps UI is not compatible with custom metadata any longer.
 To use custom type mappings you manually specify, you need to clone and run a compatible version yourself.
 
 The last release including the custom types in the UI is [`v0.96.1` available here](https://github.com/polkadot-js/apps/releases/tag/v0.96.1)
--- a/v3/tutorials/03-permissioned-network/index.mdx
+++ b/v3/tutorials/03-permissioned-network/index.mdx
@@ -463,6 +463,9 @@ stored in the `nodeAuthorization` pallet, `wellKnownNodes` storage. You should b
 able to see the peer ids of Alice and Bob's nodes, prefixed with `0x` to show its
 bytes in hex format.
 
+> NOTE: if you are using an older node that **does not include the `scale-info` pallet, you will need to use an **older version of the UI that you build and run yourself, as custom type mappings that you needed before are no longer supported (this is a very good thing!)
+> For how to use an older apps UI, see the [instructions at the end of this page](#apps-ui-use-without-scale-info).
+
 We can also check the owner of one node by querying the storage `owners` with the
 peer id of the node as input, you should get the account address of the owner.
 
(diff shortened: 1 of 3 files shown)


sacha-l:
@NukeManDan : I think that having both paths is comprehensive and a nice ting to have, but I think we should just specify in the "Before you begin" section that this tutorial expects using a node template that includes the `scale-info` pallet. As a general rule, we should always try and provide a single path forward. WDYT @lsgunnlsgunn ?

lsgunnlsgunn:
Yes, a single path in anything procedural or instructional. When you talk about a subject (informationally, conceptually, it is fine to mention that there's more than one way to skin the cat, but, in general, opt for a single path and in most cases not the "shortcut" path like keystroke combinations).

sacha-l:
@NukeManDan : An idea came to mind: let's use this content and put it in the "Troubleshoot common issues" section of our new docs. I'm sure folks will run into needing to know this at some point. WDYT?

v3/tutorials/03-permissioned-network/index.mdx
```diff
@@ -614,6 +593,47 @@ You are at the end of this tutorial and are already learned about how to build a
 permissioned network. You can also play with other dispatchable calls like
 `remove_well_known_node`, `remove_connections`.
 
+## Apps UI use without `scale-info`
+
+> The instructions bellow _only_ apply to nodes that do not use the `scale-info` pallet that all modern Substrate nodes do.
+> This powerful and very useful feature was introduced [in this PR](https://github.com/paritytech/substrate/pull/8615).
+> We highly recommend including this in your node!
+
+The hosted apps UI is _not compatible_ with custom metadata any longer.
```
Eve-Parity, line 602:
```suggestion
The hosted apps UI is not compatible with custom metadata any longer.
```

Eve-Parity (approved):
Generally speaking, notes, cautions, and warnings usually take the following format (only one `>` is necessary - maybe the auto formatting kicked in):
>**NOTE:** f;alsjiwalalajs;goijas;ofj;s;lfjfj
dshfa;sfjasdjf;alsjf;asjf;aj
ldskjfal;sfj;sjf

sacha-l:
@NukeManDan @lsgunnlsgunn @Eve-Parity : I don't think this should get merged.

NukeManDan:
what is missing @sacha-l ? users downstream are hurting because this is not mentioned. we need it yk reference for support asap in some form

sacha-l:
The content you're adding is out of context: the tutorial should maintain a "single path approach". In #607 I included a [commit](https://github.com/substrate-developer-hub/substrate-docs/pull/607/commits/e617b568b2204c2d7b9484b59ddbabcaf4f36f2d) that removes the step for adding custom types (also ensuring that other tutorials aren't misleading in that way). If anything (as suggested above), the explanation should be brief and included in the "Before you begin" section (or some other part of our docs that caters to "conceptual" or "explanatory" types of content).

New comers doing our tutorials won't be juggling different versions of Polkadot JS apps. So I don't think its necessary to show them how to go back a version &mdash; at least not here. And even though - we should always be pushing folks to using the latest (and factor in our maintenance costs).

lsgunnlsgunn:
I agree with @sacha-l this feels out of place. Can you create a separate standalone topic that covers the information you want to capture and we'll find a better home for it.

lsgunnlsgunn:
Tutorials are not the right place to put odds and ends of content. I see a lot of that kind of thing (most often these bits belong under the reference heading but we could collect a bunch of topics about polkadot JS as a starting point).

NukeManDan:
while I see your point Lisa, this has been a issue for support that needs a hone and needs to be mentioned somehow in the docs. this week alone I have had no less than 5 users ask me thus question and they should have a solution to the problem available in the docs...

perhaps this is a good division of where I should work with our support team to create stack overflow and/or a dedicated trouble shooting *page* that we can deposit this content. We still absolutely need to mention thus in the docs, even if just a line or two on what the error they see may be and a link to such content. 

We should resolve his to move forward on thus ASAP as a team in general 🙏

NukeManDan:
#478  and #477 solves these issues medium term. We should expose this in a separate PR that is versioned to use the right version of the apps UI -- moving to here: https://github.com/substrate-developer-hub/substrate-docs/issues/667

PR #608, opened 2021-11-24
Minimalistic runtime

shibshib:
This is a guide to creating a minimalistic runtime for forkless chain upgrades in cases where the block length is too small.

### DIFF
--- a/static/assets/contribute-templates/how-to-template.md
+++ b/static/assets/contribute-templates/how-to-template.md
@@ -1,36 +1,624 @@
-# Guide title [answers "How do I ...?]
+# My block length is too small to do forkless runtime upgrades, what do I do?
 
-Replace this paragraph with an overview that provides a brief summary of what this guide is about and why it's useful.
-The overview section can be more than one paragraph. It does not require any heading before the text.
+As you may be aware at this point, runtime upgrades must always abide by the block time requirements -- regardless of running it as SUDO or not. For example, if your block length is 100KB and your runtime WASM is 120KB, then you will not be able to perform a chain upgrade as it is too large. 
 
-## Use cases [optional - if it is redundant, leave it out]
+At Bittensor, we have run into an extreme version of this problem: we accidentally shrunk our block length to 65KB! This effectively made it impossible to perform chain upgrades as our WASM files were at least 300KB in size. 
 
-Remove this section if it repeats the guide title or information covered in the overview section. Include this section if it adds value. For example, if a single use case requires more explanation that provided in the guide title or in the overview, use this section to provide the more detailed explanation.
+## Use cases
 
-## Before you begin [required if there are prerequisites]
+This is particularly important if you've accidentally shrunk your block length too much and need to bring your chain back to an upgrade-able state. 
 
-Replace this template content with an introductory sentence and list of prerequisites similar to the following.
+## Before you begin 
 
-Before you begin, verify the following requirements:
+Before you begin, you should have some knowledge of Forkless Runtime upgrades, you can find a guide on this here: https://docs.substrate.io/tutorials/v3/forkless-upgrades/.
 
-- What should someone **have** before reading this article?
 
-- What should someone **know** before reading this article?
+## How do I recover my chain?
 
-- What should someone **do** before reading this article?
+We will use our experience at Bittensor as an example for this how-to guide. What we had done was accidentally shrink our block length to 65KB in an effort to over-optimize. This made it impossible to run forkless chain upgrades as the WASM files we were uploading were far too large for the blocks at 300KB. As a result, we had to "thread the needle". That is, we created a very small runtime (64.8KB) that consisted of only the ability to perform a chain upgrade with a larger block length, then followed it with another upgrade that restored the code back to its working state. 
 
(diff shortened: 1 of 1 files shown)


sacha-l:
@shibshib: This is awesome, thanks for contributing ! Will take a closer look over the next few days. I'll also make the changes to have it render with our Gatsby setup. Then it should be good to go. At first glance it looks thorough and well articulated. Well done.

P.S.: This guide makes you sound like a survivor from a runtime rescue mission 😆 ! Thanks for sharing your experience, its super valuable.

shibshib:
Thank you @sacha-l ! And it was absolutely a rescue mission, because we would have had to restart the chain otherwise :D

static/assets/contribute-templates/how-to-template.md
```diff
@@ -1,36 +1,624 @@
-# Guide title [answers "How do I ...?]
+# My block length is too small to do forkless runtime upgrades, what do I do?
 
-Replace this paragraph with an overview that provides a brief summary of what this guide is about and why it's useful.
-The overview section can be more than one paragraph. It does not require any heading before the text.
+As you may be aware at this point, runtime upgrades must always abide by the block time requirements -- regardless of running it as SUDO or not. For example, if your block length is 100KB and your runtime WASM is 120KB, then you will not be able to perform a chain upgrade as it is too large.
```
kianenigma, line 3: I can easily think of tweaking `frame_system` and its signed extensions to allow only certain transactions to go over the block length limit. This will ensure that something like `set_code` will never exhaust the block resources. 

There should be warnings that doing this kind of thing is a bit risky.

shibshib: This kind of tweak would be **tremendously** helpful, as we were surprised that we could not override the block length limit using Sudo.

I agree with you about warnings, added them in to the article..

static/assets/contribute-templates/how-to-template.md
```diff
@@ -1,36 +1,631 @@
...
-1. Open ...
+It's important to prepare both runtimes and push them consecutively. First prepare the normal runtime you are trying to upgrade with and save it aside for use in step 10. Now, let's create this minimalistic runtime. It must be smaller than your present block length. In our case, the block length was 65KB. All of the following changes occur in `runtime/src/lib.rs`.
 
-## Examples [required - link to at least one example]
 
-Replace this paragraph with an introductory sentence followed by a link to at least one working example that corresponds with the information covered in this guide.
+1. Remove any and all custom code you added to your runtime. This means removing the implementation from `construct_runtime!`, all associated `parameter_types!`, and all implementations for runtime.
 
-## Related resources [optional - add links as needed]
+2. Comment out some unneeded configs and functions. Things like `native_version()`, `frame_benchmarking`, and `pallet_transaction_payment_rpc_runtime_api` can all be removed.
```
kianenigma, line 32: `frame_benchmarking` is not part of the final wasm, unless if you are mistakenly enabling its feature. Under correct circumstances, removing it should not affect the wasm size at all.

static/assets/contribute-templates/how-to-template.md
```diff
@@ -1,36 +1,631 @@
...
+5. Remove all unnecessary `signedExtensions`. 
+  ```rust
+  pub type SignedExtra = (
+    frame_system::CheckSpecVersion<Runtime>,
+    //frame_system::CheckTxVersion<Runtime>,
+    // frame_system::CheckGenesis<Runtime>,
+    // frame_system::CheckEra<Runtime>,
+    // frame_system::CheckNonce<Runtime>,
+    // frame_system::CheckWeight<Runtime>,
+    // pallet_transaction_payment::ChargeTransactionPayment<Runtime>
```
kianenigma, line 75: might be worth mentioning that removing some of these makes your chain be easily attackable. I would personally NOT remove these in your scenario. You need some of these to keep the chain safe. 

For example, not checking the Nonce is basically suicide if any pallet is active to end-users. 

Or not checking the Genesis hash makes you vulnerable to long-range attacks.

kianenigma (commented):
Interesting story! Thanks for sharing it. 

All in all, I would also add that I think if you want to strip down your runtime, you probably do NOT want to remove system pallet and its signed extensions. You can and should remove all other pallets, do the maintenance needed, then bring them back.

static/assets/contribute-templates/how-to-template.md
shibshib: @kianenigma Agree with you it's rather dangerous. However in the article we do mention that this minimalist runtime should be followed **immediately** by another upgrade that will bring back your original code. I will add an update to the article to mention your points.

static/assets/contribute-templates/how-to-template.md
shibshib: Interesting, I'll add this in.

NukeManDan:
Hey @shibshib - the `main` branch has moved, and now we would want to ask @lisa-parity if this should live in the new docs. If you think so please open a new PR targeting the new docs site? Or at least an issue to flag that we should bring this in as new content :pray: 

cc @kianenigma @sacha-l

PR #625, opened 2021-11-28
Bring `Config` of `Balances` up to date with solution

maltekliemann:
The tutorial of _Add a Pallet_ refers to the `Config` of the `Balances` pallet, but is missing the definition of `MaxReserves` and `ReserveIdentifier`, which are contained in the solution for the tutorial. The members of the trait are also reordered to line up with the solution.


sacha-l (approved):
Thanks @maltekliemann - appreciate the gesture ! 🙏🏻

PR #587, opened 2021-11-21
Bugfix: remove #[pallet::generate_storage_info]

mrquincle:
The code created by following the tutorial does not compile:

```
 cargo check -p node-template-runtime   Compiling node-template-runtime v3.0.0-monthly-2021-10 ($HOME/workspace/blockchain/substrate-node-template/runtime)
    Checking pallet-template v3.0.0-monthly-2021-10 ($HOME/workspace/blockchain/substrate-node-template/pallets/template)
error[E0277]: the trait bound `Vec<u8>: MaxEncodedLen` is not satisfied
  --> pallets/template/src/lib.rs:43:11
   |
43 | #[pallet::generate_storage_info]
   |           ^^^^^^^^^^^^^^^^^^^^^ the trait `MaxEncodedLen` is not implemented for `Vec<u8>`
   |
   = note: required because of the requirements on the impl of `StorageInfoTrait` for `frame_support::pallet_prelude::StorageMap<_GeneratedPrefixForStorageProofs<T>, frame_support::Blake2_128Concat, Vec<u8>, (<T as frame_system::Config>::AccountId, <T as frame_system::Config>::BlockNumber), frame_support::pallet_prelude::ValueQuery>`
note: required by `storage_info`
  --> $HOME/.cargo/git/checkouts/substrate-7e08433d4c370a21/bf9683e/frame/support/src/traits/storage.rs:71:2
   |
71 |     fn storage_info() -> Vec<StorageInfo>;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0277`.
error: could not compile `pallet-template` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```

Remove `#[pallet::generate_storage_info]` preceding `pub struct Pallet<T>(_);`

PR #626, opened 2021-11-29
Remove extra line of `generate_storage_info`

xiankuncheng:
Compare with [final solution](https://github.com/substrate-developer-hub/substrate-node-template/blob/tutorials/solutions/proof-of-existence/pallets/template/src/lib.rs), there isn't this line of code exists. And it will cause "the trait `MaxEncodedLen` is not implemented for `Vec<u8>`" error thrown and stop the cargo check process.


xiankuncheng:
Closed since just realized there already has a [PR for fixing this](https://github.com/substrate-developer-hub/substrate-docs/pull/587) raised.

PR #628, opened 2021-11-29
[WIP] - ServiceWorker Cache Issue

### LINKED ISSUE #578: Empty page on Kitties Tutorial

_No response_

Got an empty page when visiting https://docs.substrate.io/tutorials/v3/kitties/pt1/
with the following error in the console:
```
Uncaught TypeError: can't access property "map", n is undefined
    l DocsNav.tsx:78
    React 7
    unstable_runWithPriority scheduler.production.min.js:18
    React 6
    unlisten index.js:144
DocsNav.tsx:78:15
```

Browser:
Firefox Developer Edition 95.0b8 (64-Bit), Ubuntu 20.04.3 LTS

Goes away after hard reload (ctrl + shift + r).

imadarai:
This is just a temp fix to check for updated content and force a hard-reload using `gatsby-browser` API.  
Resolves #578

### DIFF
--- a/gatsby-browser.js
+++ b/gatsby-browser.js
@@ -31,6 +31,10 @@ export const wrapPageElement = ({ element }) => (
 
 export const onRouteUpdate = ({ location }) => scrollToAnchor(location)
 
+export const onServiceWorkerUpdateFound = () => {
+  window.location.reload(true)
+}
+
 /**
  *
  * @desc - a function to jump to the correct scroll position


gatsby-browser.js
```diff
@@ -31,6 +31,10 @@ export const wrapPageElement = ({ element }) => (
 
 export const onRouteUpdate = ({ location }) => scrollToAnchor(location)
 
+export const onServiceWorkerUpdateFound = () => {
+  window.location.reload(true)
+}
+
```
rmnprkrl, line 37: I would not go this direction, it could potentially cause very strange behaviour.
Better to isolate the problem and solve it. I've added some comments to notion.

user-09876: What about:

```
if (window.confirm('This app has updated, please reload the page')) {
  window.location.reload(true)
}
```

So it leaves a choice to the user, to:
- reload or not the page (if `okay` is clicked at confirmed) and get the updates
- alerts them that some updates are available, and that some strange behavior might happens if they don't reaload (if `no` is clicked at confirm).

rmnprkrl:
service worker deregistered: https://github.com/substrate-developer-hub/substrate-docs/pull/698

rmnprkrl:
As this is not a preferred solution I would close this PR.

PR #646, opened 2021-12-02
Adding close button to banner

### LINKED ISSUE #612: Add a close button on the Substrate StackExchange msg overlay

Chrome

Can we add a close button on the Substrate StackExchange msg overlay on Substrate? It is pretty annoying and couldn't close it. Yes I know scrolling the screen down will make it disappear.

Also, once closed, I would suggest make the message don't popup again for that session. (So it is fine to have the overlay appears again when the window/tab is closed and reopen again).

Goto https://docs.substrate.io

imadarai:
Closes #612 

Added a session-storage hook, that keeps track of the banner.  If a user closes the banner, the value is updated in the sessionStorage and it doesn't re-appear, until another session is opened.  Banner will also hide, if user scrolls down. 

If the approach is good, we can replicate it on `substrate.io` as well.

### DIFF
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -2,12 +2,14 @@ import cx from 'classnames'
 import React, { useEffect, useState } from 'react'
 
 import useScrollListener from './Hooks/use-scroll-listener'
+import useSessionStorage from './Hooks/use-sessions-storage'
 import Icon from './Icon'
 import Link from './Link'
 
 const Banner = () => {
   const scroll = useScrollListener()
   const [isScrolled, setIsScrolled] = useState(false)
+  const [isBannerOpen, setIsBannerOpen] = useSessionStorage('banner', true)
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


imadarai:
Found a bug while running `yarn build`.  Working on a fix.

imadarai:
Ready for review.

NukeManDan:
Is it possible to have the scroll behavior removed? If the close ability is there, IMHO it's fine to leave it up untill it's closed.

Will this also be shown again for _every_ new page a user navigates to and on page reload?
A nice-to-have: a timeout site wide of say 5 minutes (easy to change const somewhere) before reloads or page changes would bring it back up.

NukeManDan:
Closes #612 
(this keyword links the issue in the comment cc @imadarai so it closes automaticaly)

rmnprkrl:
This will be replaced with WIP banner feature this week, where banner is generated from `.md` file so contributors can maintain the content.

PR #655, opened 2021-12-04
Fix false instruction in git repo check out

sicongzhao:
We need first cd into the folder and then use 'git checkout' command


NukeManDan (approved):
Thanks for the spot!

sicongzhao:
Polkadot address: 15Vu9wk1AUzxVHT5p1GEYEk819sHqty8gWd13ZMngj9fZXRQ

PR #663, opened 2021-12-06
Docs: remove indent in paragraphs

angelolovatto:
Removes indentation of the last two paragraphs of the "Install Rust and the Rust toolchain" section.
This was causing the text to show in monospace font and broke a hyperlink.

Screen Shot 2021-12-06 at 20 02 15


NukeManDan:
Thanks for the report and this fix! actually it needed to be 3 spaces for a continued list in this case - fix in #666 :grin:

PR #664, opened 2021-12-07
Organise gatsby node

rmnprkrl:
- [x] separate node redirects config
- [x] create node pages revamp
- [x] use full names for page templates 
- [x] separate Lunr config in node-resolver (separate index build)

Check on (Will be migrated into new infrastructure based on Mkdocs):
+ moving from `/v3` to `/content`
+ removing indexes from folders and use folders to generate slugs
+ remove version prefix from frontmatter slug
+ fix `section` field in frontmatter (kebab case or similar sanitized string formatting)


imadarai (approved):
Looks good to me.  
I cloned it locally and tested for any un-intended behavior.  All pages checks out + redirects and search.

PR #660, opened 2021-12-06
Cumulus content updates for `polkadot v0.9.13`

NukeManDan:
Closes #524, #513

The launch content is blocked on #634 though, and is presently hidden from nav, as the content is incorrect.  
Will correct this in a follow up PR.

PR #671, opened 2021-12-08
add step for chainspec creation (howto)

### LINKED ISSUE #513: Parachain connect should mention the creation of a chain spec
### Content request

https://docs.substrate.io/how-to-guides/v3/parachains/connect

We should add as step 2 a detailed description on how to create a proper chain spec. Beginning with a json chain spec, where the user again ensures that all values are as expected, e.g. the parachain id. Then the creation of the raw chain spec out of the json chain spec. In the next steps we should then also use this created chain spec.

### Are you willing to help with this request?

Yes!

bernardoaraujor:
close #513


NukeManDan:
@bernardoaraujor THANK YOU for this, although I would rather have you review #660  that includes more detail :pray:

PR #682, opened 2021-12-13
Add back playground-hero.png image and make frontmatter `version` and `keywords` consistent

jimmychu0807:
- Add back `playground-hero.png`. It has been missing this image. This solves a warning.
- Frontmatter `version` are all converted to string, `3.0` is quoted to be `'3.0'`. This solves a warning.
- Frontmatter `keywords` are all covered to array of string. This solves a warning.

PR #696, opened 2021-12-14
Add missing playground img

rmnprkrl:
Notice playground image is missing on homepage for some reason, adding back


rmnprkrl:
Noticed @jimmychu0807 is fixing this here: https://github.com/substrate-developer-hub/substrate-docs/pull/682

PR #709, opened 2021-12-20
Update index.mdx misspellings

mjluck:
Updated what I believe are misspellings as I followed the guide.

On lines 513 to 516, the final sentence is confusingly worded to me but I am not sure what the original author intended for it. "If running node in a public internet, you need to otherwise put the reachable nodes as bootnodes of the Chain Spec.", sounds better to me but I might not understand what the statement is referencing correctly yet.

Got everything running though, great tutorial so thanks to everyone!


NukeManDan (approved):
spellz is hardz.

Thank you!

PR #713, opened 2021-12-23
rewrapped descriptions so work better on small screens

gilescope:
On my mac laptop/firefox the descriptions were not wrapping themselves (I think they're rendered preformatted) and thus you had to scroll right to read what they were saying. It should be easier to see everything without having to scroll right now.

*No changes to any text.*


gilescope:
let me raise this against develop

PR #715, opened 2021-12-24
Fix broken links in how-to-guides

gilescope:
Links are broken as the leading spaces treated it as a `pre` block rather than interpreting the links.
Found `parachains / connect to relay chain` had fallen off (had same starting letter as another) so put it back.

I put back `01-basics / e-custom-chain-spec` which had also fallen off, but locally it _still_ doesn't seem to appear in the sidebar. - Any ideas? (We're no worse off than before, but still would be nice to figure out why it's not appearing)


v3/how-to-guides/01-basics/e-custom-chainspec/index.mdx
```diff
@@ -13,33 +11,35 @@ keywords:
   - private
   - network
   - bootstrap
+version: '3.0'
```
NukeManDan, line 14:
```suggestion
version: 'polkadot-0.9.13'
```

NukeManDan (approved):
LGTM

src/components/DevNavMenu.tsx
```diff
@@ -279,6 +279,10 @@ const DevNavMenu = {
           {
             title: 'Genesis Configuration',
             link: '/how-to-guides/v3/basics/genesis/',
+          },
+		  {
+            title: 'Custom Chain Specifications',
+            link: '/how-to-guides/v3/basics/genesis/',
```
NukeManDan, line 285: @gilescope this is what is needed to expose things in the nav (this file in general)

PR #733, opened 2022-01-05
fix storage getter name

TarekkMA:
In the rest of the tutorial the getter `all_kitties_count` is not used


sacha-l (approved):
This is correct, thank you.

PR #744, opened 2022-01-08
Fix "sucessfully" typo

cor:


sacha-l (approved):
Thanks for your PR @cor! 🤠

PR #745, opened 2022-01-10
[docs] fix typo in commands (tutorials/05-private-network)

tokiedokie:
The command could not be executed, so I fixed it.


jimmychu0807 (approved):
@tokiedokie, thanks for the contribution.

PR #746, opened 2022-01-10
Fix error in example code of runtime storage documentation

sudipghimire533:
In the given example, generic parameter was specified without restricting but later on uses type parameter of that generic. This have generated error regarding associated type not found. Restricting this generic type however resolved the error.


sacha-l:
Thanks @sudipghimire533 , looks good to me!

sacha-l (approved):

PR #752, opened 2022-01-12
Fix Ubuntu install command (tutorial)

Chrecci:
The command in the tutorial doesn't match the one in the actual setup docs: https://github.com/substrate-developer-hub/substrate-node-template/blob/main/docs/rust-setup.md . Leads to all sorts of install/compilation errors


NukeManDan:
updated upstream

PR #753, opened 2022-01-12
Newsletter subscription block in Footer

je-boska:


.gitignore
```diff
@@ -74,3 +74,6 @@ yarn-error.log
 !.yarn/versions
 # Yarn Integrity file
 .yarn-integrity
+
+# Local Netlify folder
```
rmnprkrl, line 78: this folder needs to be tracked by remote

rmnprkrl: removing this would break deploy setup (atm redirects)

src/components/site/Newsletter.js
```diff
@@ -0,0 +1,162 @@
...
+    setLoading(true)
+    const data = new FormData(event.target)
+    fetch('https://paritytechnologies.activehosted.com/proc.php', {
+      method: 'POST',
+      body: data,
+      mode: 'no-cors',
+    })
+      .then(response => {
+        setLoading(false)
+        console.log(response)
```
rmnprkrl, line 38: remove the response in console

PR #739, opened 2022-01-06
Replace seed with secret phrase

lisa-parity:

PR #756, opened 2022-01-14
Fix #755

### LINKED ISSUE #755: [Start a private network tutorial] Does the `key inspect` command really require a password?
https://docs.substrate.io/tutorials/v3/private-network/#generate-local-keys-using-the-node-template
In the sixth step of this part, I enter the correct password and the wrong password, and even press Enter without entering the password, the output results are the same. So I think `--password-interactive` can be removed here, no need to enter a password.
I read the command description [here](https://docs.substrate.io/v3/tools/subkey/#inspecting-keys) and still don't think a password is required.

04637:
remove `--password-interactive` in "Generate local keys using the node template" Step 6


04637:
After merging #739, the problem is gone.

PR #763, opened 2022-01-17
fix yarn.lock mismatch

rmnprkrl:
This commit: https://github.com/substrate-developer-hub/substrate-docs/commit/cd49ded85b5594d8013f5ad017026cd8cdda9325 introduced local `yarn.lock` edit which causes CI builds to fail from this commit up

This PR provides regenerated `yarn.lock` which should fix the issue


rmnprkrl:
@sacha-l ^^

sacha-l (approved):

PR #764, opened 2022-01-18
Fixed correct method calls

452MJ:
Fix `Unhandled Rejection (TypeError): api.query.substrateKitties.multi is not a function`.


452MJ:
@sacha-l Please review util you are free

sacha-l (approved):
This is correct, thanks for the fix.

PR #777, opened 2022-01-19
Add extrinsic_index to gen_dna function to resolve issue of hash dupl…

cc-michael-x:
…ication when transactions are in the same block


sacha-l (approved):

sacha-l:
Merging this as an improvement. However, the tutorial write-up and helper files needs reworking based on the latest [solution](https://github.com/substrate-developer-hub/substrate-node-template/blob/tutorials/solutions/kitties/pallets/kitties/src/lib.rs). Follow-up issue: #885.

PR #791, opened 2022-01-21
Document Rococo parachain slot assignment

### LINKED ISSUE #665: Document permissionless Rococo slot aquisition
Internal notes on flow
https://github.com/paritytech/delivery-services/issues/85#issuecomment-923780598

Issue https://github.com/paritytech/polkadot/issues/3942 / PR https://github.com/paritytech/polkadot/pull/3943 

Wait to publish formally on opening of this for rococo.
- [ ]  Move to a How-to-guide instead of the tutorial page https://docs.substrate.io/tutorials/v3/cumulus/rococo/
- [ ] Parse out HTG about non-sudo slot managment
- [ ] #241 
- [ ] https://github.com/paritytech/cumulus/pull/869
- [ ] slot details https://github.com/paritytech/polkadot/pull/4577/files

stiiifff:
Removed section about parachain slot auction & crowdloan.
Added description of permanent or temporary parachain slot assignment mechanism.
Close #665

/cc @NukeManDan @sacha-l let me know if I need to refine some parts.


sacha-l (commented):
Thanks for putting this together @stiiifff! Just added my review as a commit - it was easier to do as it's mostly reformatting stuff from before. I think we should turn the last part into some action items instead of giving a description of the two slot types and leaving it at that.

Is there an extrinsic to call for requesting a permanent or temporary slot? What happens after that? Can a team keep requesting a slot even after their 3 day period expires?

Also, is it still 5 ROCs to reserve a para id?

stiiifff:
@sacha-l the `assign_perm_parachain_slot` and `assign_temp_parachain_slot` extrinsics of the `assigned_slots` pallet are used to register the permanent and temporary slots respectively. Their access is restricted by the configurable `AssignSlotOrigin`, which is currently set to the root origin (see [here](https://github.com/paritytech/polkadot/blob/1fdeef8eb73a7b93c4913818f670563f89daa321/runtime/rococo/src/lib.rs#L675)). This could be changed at a later point to a governance origin.

After a parachain's temporary slot lease has ended, it is automatically downgraded back to a parathread. It will remain a parathread for at least the next slot lease, and then, for the slot lease after that, a new batch of parathreads will be upgraded to parachains as per the logic describe [here](https://github.com/paritytech/polkadot/blob/1fdeef8eb73a7b93c4913818f670563f89daa321/runtime/common/src/assigned_slots.rs#L391). Do you think we should further detail this behavior in the doc ?

The reservation of a ParaId still works as before AFAIK.

NukeManDan (approved):
I took this a LOT further than the original purpose, as we needed to preserve:
- Crowdloan information in a new HTG
- Clean up of duplicated HTGs

I added a few extra notes on Rococo, as at the moment you _must_ contact Parity to get a slot at all. Not ideal, but what we have for now.

NukeManDan:
Also adding a note for those controlling a sudo key on the admin side from @stiiifff :bow: 

---

Here is the flow:

- First thing is to initialize a para with `parasSudoWrapper.sudoScheduleParaInitialize`) and wait for it to be onboarded as a parathread.
- Assign a temporary parachain slot with `parasSudoWrapper.sudoAssignTempParachainSlot`, specifying the para Id and lease period start (Current lease period or next one).
image
- This will create a lease and the parathread will be upgraded to a parachain when its start lease period begins.
image
- When a slot's lease ends, the parachain is automatically downgraded to a parathread.
- `parasSudoWrapper` pallet keeps track of temporary slots so it can distribute turns fairly between temp slots.
image
- You can also assign a permanent (or long-running) slot with `parasSudoWrapper.sudoAssignPermParachainSlot` or unassign a slot (permanent or temporary) with `parasSudoWrapper.sudoUnassignParachainSlot` (this clears any lease which downgrades the para back to a parathread).
- This all relies on the existing `slots` pallet & lease mechanism, nothing fancy.
- The following parameters need to be to configured (let me know what you think should be the values for Rococo):
```rust
pub const LeasePeriod: BlockNumber = 7 * DAYS;
// The number of lease periods a permanent parachain slot lasts
// (multiplied by LeasePeriod to get the slot duration in blocks)
pub const PermanentSlotLeasePeriodLength: u32 = 26; 
// The number of lease periods a temporary parachain slot lasts
// (multiplied by LeasePeriod to get the slot duration in blocks)
pub const TemporarySlotLeasePeriodLength: u32 = 1;
// The max number of permanent slots that can be registered
pub const MaxPermanentSlots: u32 = 5;
// The max number of temporary slots to be scheduled per lease periods
pub const MaxTemporarySlotPerLeasePeriod: u32 = 10;
```

NukeManDan:
#801 should land _before_ this. Should be ready to merge as soon as Polkadot releases v0.9.16 - ETA Monday 2022-02-01 !

stiiifff:
@NukeManDan re. the sudo section, the calls do not go through `parasSudoWrapper` anymore. This was the case in an early iteration of the design, but then I migrated the logic in the `assignedSlots` pallet. I feel this additional section does not really add meaningful info (and  risk making things more confusing than intended), so I'd skip it entirely for now, what you think ?

v3/tutorials/09-cumulus/d-rococo/index.mdx
```diff
@@ -25,162 +25,127 @@ keywords:
...
+- **Maximum number of temporary slots**: up to 20 temporary slots
+- **Maximum temporary slots allocated per leased period**: up to 5 temporary slots per 3-day temporary lease periods
 
-![parachain-bid.png](../../../../src/images/tutorials/09-cumulus/parachain-bid.png)
+These are subject to change, based on the needs of the community.
 
-### Crowdloans
+🎉**_Congratulations!_**🎉  
+With your slot activated by the Parity team, you're now able to test your parachain on the Rococo test net!
+Please do note that when your temporary slot's lease ends, the parachain is automatically downgraded to a parathread. Registered and approved slots are cycled through automatically in a round-robin fashion, so you will expect to come back online as a parachain from time to time.
```
sacha-l, line 112:
```suggestion
Please do note that when your temporary slot's lease ends, the parachain is automatically downgraded to a parathread. 
Registered and approved slots are cycled through automatically in a round-robin fashion, so you will expect to come back online as a parachain from time to time.
```
I thought that once your temporary slot is up and you're downgraded to a parathread, you need to re-register for a temporary slot again. ".. you will expect to come back online as a parachain from time to time" makes it sound like you could just be reassigned a slot automatically even after your slot period runs out. Is this the case?

sacha-l: cc @NukeManDan.

NukeManDan:
> @NukeManDan re. the sudo section, the calls do not go through parasSudoWrapper anymore. This was the case in an early iteration of the design, but then I migrated the logic into the assignedSlots pallet. I feel this additional section does not really add meaningful info (and risk making things more confusing than intended), so I'd skip it entirely for now, what you think ?

Removed. Anything we _should_ state to help those understand the manual process that is needed on our end (if any) @stiiifff ? It is unclear to me what Parity/ the sudo key holder must do for someone to acquire a slot, and for that slot to automatically be renewed in a round robin style.

v3/tutorials/09-cumulus/d-rococo/index.mdx
stiiifff: @NukeManDan @sacha-l so, when a para temp slot's lease ends, the parachain is downgraded back to a parathread (this is done by the `slots` palet's `manage_lease_period_start` function). A new batch of parathreads are then upgraded to parachains for the duration of a temp slot lease (this is done by the `assigned_slots` pallet's `allocate_temporary_slot_leases` function). The catch is that a temp slot does not expire (once registered, it remains registered until one calls `unassign_parachain_slot`), but it has multiple slot leases (in the `slots` pallet's terminology) that have a fixed duration (defined by the `TemporarySlotLeasePeriodLength` param). Does that make sense ? 😅

stiiifff:
@NukeManDan The only operation that needs to be done manually is to call either `assign_perm_parachain_slot` or `assign_temp_parachain_slot` with the para ID. Having this become governance-drive is just a matter of changing the `assigned_slots` pallet's `AssignSlotOrigin` parameter.

v3/tutorials/09-cumulus/d-rococo/index.mdx
sacha-l: Makes sense! OK. That would imply that the more registered paraID's, the least likely a newly added paraID will be chosen for the next available slot? Is there a cap to the number of para IDs that can be registered?

PR #840, opened 2022-02-19
docs(TemplateModule): fix inconsistencies in code

robertohuertasm:
This PR fixes some inconsistencies I found while doing the tutorial as the code wasn't working. I had to change the way to access the `api` and the `currentPair` in order to make it work.


NukeManDan:
Thanks for this! This is fixed in #861 as part of a larger refactor.

PR #852, opened 2022-02-23
Fixing print macro in snippets

### LINKED ISSUE #838: Debug `print` function should be `print!`

Other (please write below)
Brave

I think all examples that use `print` function at https://docs.substrate.io/v3/runtime/debugging/ should be using `print!` (with a bang character) instead

In the Substrate Debugging deocumentation here https://docs.substrate.io/v3/runtime/debugging/

It says that we can use Substrate's `print` function with
```
print("My Test Message");
```

But if I do that I get error
```
error[E0423]: expected function, found macro `print`
  --> node/src/chain_spec.rs:51:5
   |
51 |     print("My Test Message");
   |     ^^^^^ not a function
```

Whereas if I use a bang character by changing it to `print!("My Test Message");` then it works

tarrball:
Closes #838 

I searched the docs and only found problems with this one file.

### DIFF
--- a/v3/docs/03-runtime/j-debugging/index.mdx
+++ b/v3/docs/03-runtime/j-debugging/index.mdx
@@ -77,7 +77,7 @@ pub fn cause_error(origin) -> dispatch::DispatchResult {
 	// Check it was signed and get the signer. See also: ensure_root and ensure_none
 	let _who = ensure_signed(origin)?;
 
-	print("My Test Message");
+	print!("My Test Message");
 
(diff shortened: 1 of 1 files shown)


NukeManDan (approved):

NukeManDan:
@tarrball thanks! please always target `main` - a new PR?

NukeManDan:
https://github.com/substrate-developer-hub/substrate-docs/pull/869

PR #850, opened 2022-02-22
--json is the new --raw

gilescope:
As of https://github.com/paritytech/substrate/pull/10771 `--json` is the new `--raw`

PR #866, opened 2022-03-02
Update live private network with changes

lisa-parity:
Updating the live version with some content changes. I got it working all the way through finalization. The live version has both the predefined account keys and the generated keys, so not as tidy as the new version in progress in the `main-md `branch.

Change --raw to --json per:
https://github.com/substrate-developer-hub/substrate-docs/pull/850

### DIFF
--- a/v3/tutorials/05-private-network/index.mdx
+++ b/v3/tutorials/05-private-network/index.mdx
@@ -21,59 +21,301 @@ relevantSkills:
 
 This tutorial illustrates how you can start a small, standalone blockchain network with an **authority set** of private **validators**.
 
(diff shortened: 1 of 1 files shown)


NukeManDan:
This is based on https://github.com/paritytech/substrate/pull/10771 ? So far the templates do not use this, but will in 0.9.18, so we can test then :) 

This is also effect the cumulus content that used that flag, not sure if this has made it to there yet.

lisa-parity:
Per @NukeManDan comment, the --raw is restored and 0.9.18 has been released, so we should be okay to merge this now.

lisa-parity:
Per @NukeManDan comment, the --raw is restored and 0.9.18 has been released, so we should be okay to merge this now.

PR #875, opened 2022-03-05
Fix markdown list style in mint token page.

xiankuncheng:


sacha-l (approved):
Thanks for fixing this @xiankuncheng !

PR #895, opened 2022-03-15
fix a word

hzy1919:


NukeManDan:
No, this is the intended word. Although I could see an argument for either working.

hzy1919:
> No, this is the intended word. Although I could see an argument for either working.

ok.

Issue #884, opened 2022-03-10
PoE Tutorial to compile from step to step


NukeManDan:
> I would still prefer we start with a template which compiles from step to step. this could help someone realize there is an issue before getting too far into the tutorial 
> _Originally posted by @shawntabrizi in https://github.com/substrate-developer-hub/substrate-docs/pull/861#pullrequestreview-905083794_

This should not be too hard to implement, and the solution could be constructed with each commit for these steps. 
Possibly using a different base skeleton for the pallet template: https://sacha-l.github.io/collectables-ethdenver-workshop/#/setup

lisa-parity:
Probably superseded by:
https://github.com/substrate-developer-hub/substrate-docs/pull/1352

PR #910, opened 2022-03-22
[PoE] Adding initial template that compiles.

arturgontijo:
This PR solves (#884) by adding a pallet template (`lib.rs`) that compiles and also has all the `TODO`s that will be updated by the next steps in the tutorial.

@NukeManDan let me know if I got the idea correctly.


lisa-parity (approved):
LGTM

PR #925, opened 2022-03-28
update pallet attr documentation

gilescope:
`#[pallet::generate_storage_info]` is now defaulted.

### DIFF
--- a/v3/docs/03-runtime/b-macros/index.mdx
+++ b/v3/docs/03-runtime/b-macros/index.mdx
@@ -191,24 +191,24 @@ pub struct Pallet<T>(_);
 
 **When to use**
 
-You should not need to explicitly add this attribute.
-It is now assumed by default that all of the pallet storage items are bounded (i.e. have a fixed size).
+This attribute is now a no_op. It is now assumed by default that all of the pallet storage items are bounded (i.e. have a fixed size).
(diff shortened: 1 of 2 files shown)


NukeManDan (approved):
LGTM - perhaps a followup to remove this would be in order? When would you need `#[pallet::without_storage_info]` and only speak to this?

ggwpez:
> When would you need `#[pallet::without_storage_info]` and only speak to this?

Mostly a legacy thing to ease transition since we did not want to force developers to update their pallet.  
This allows an opt-out for developing but in the long run it is necessary for the WeightsV2. @NukeManDan

v3/docs/03-runtime/b-macros/index.mdx
```diff
@@ -192,11 +191,13 @@ pub struct Pallet<T>(_);
 
 **When to use**
 
-Highly encouraged, but optional which will help ensure all of your pallet storage items are bounded.
+You should not need to explicitly add this attribute.
```
ggwpez, line 194: The macro got removed completely, since it would be a NO-OP now.  
Maybe worth mentioning.

ggwpez (approved):

Vovke (approved):

PR #929, opened 2022-03-31
Banner feature

je-boska:
This implements the Banner with the same design as on substrate.io, but with all data localized to the Banner component.


sacha-l (approved):
LGTM :)

PR #930, opened 2022-03-31
Fix Home page scrolling issues

je-boska:
Notion: https://www.notion.so/paritytechnologies/Docs-Home-horizontal-scrolling-dc5847c3e92742749cfb6dd5a726db66

### DIFF
--- a/src/components/Layout.tsx
+++ b/src/components/Layout.tsx
@@ -46,7 +46,7 @@ const Layout = ({ children }: any) => {
   return (
     <React.Fragment>
       <Header />
-      <main className="main-container">
+      <main className="main-container overflow-x-hidden">
         <MDXProvider components={components}>{children}</MDXProvider>
       </main>
       <Footer />
--- a/src/components/layout/homepage/DocCard.tsx
+++ b/src/components/layout/homepage/DocCard.tsx
@@ -25,7 +25,7 @@ export default function DocCard({
       <div
         className={cx(
           'w-full text-center shadow-xl bg-white p-10 md:px-6 md:py-10 xl:p-10',
-          'mb-8 lg:mb-0 m:max-w-sm hover:scale-105 transition-transform lg:w-72 xl:w-96 lg:mr-8 rounded-md',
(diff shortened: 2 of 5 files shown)


je-boska:
The only difference in the way this appears from what's currently deployed, is that the browser mockup in the Playground section now fades in from the right starting at the container width, rather than all the way from the right of the view.

NukeManDan (approved):
Trust this is fine, @je-boska is all you needed and approval? If so, please just ping on element when you need this, or request a review from me or any other admin

rmnprkrl:
@je-boska 
it is caused by this AOS transition:

image

Adding `overflow-hidden` to the wrapper like shown in the screenshot should be enough to fix it..

je-boska:
> @je-boska it is caused by this AOS transition:
> 
> image
> 
> Adding `overflow-hidden` to the wrapper like shown in the screenshot should be enough to fix it..

Yes, I added overflow-x-hidden to the wrapper of that asset.

There were fixed widths on some of the cards in the section above, which was also causing some minor horizontal scrolling. I sorted this by placing them in a grid instead. There is a visible difference in how they appear, but it's very minor.

rmnprkrl:
@je-boska could you test it doesn't break anything when added to to the main container instead?:
`<main class="main-container overflow-x-hidden">`

Adding only to that AOS wrapper as I suggested crops the transition visually..

je-boska:
I can't find any problems with adding it to the `main-container` class. Let me know if it's OK to edit the `tailwind-preset.css` file in this case.

I also fixed vertical scrolling caused by overflow on the section above ("Learn Substrate").

rmnprkrl:
Everything needs to stay in tailwind (or particular used framework in general) context, only if not possible it goes out of the system, add it to the container directly

> I can't find any problems with adding it to the main-container class

did you you test it does not break any page on any break point on VA devices?

je-boska:
Tested on mac and iOS, across Chrome/Brave, Safari and Firefox. Couldn't find any breakage.

rmnprkrl:
LGTM

PR #939, opened 2022-04-06
Fix typo in glossary

lisa-parity:


lisa-parity:
Fixed in main-md too.

Issue #931, opened 2022-04-01
Fix run_to_block example on Testing page


0xangelo:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Content request

The `run_to_block` example in the Runtime -> Testing docs calls `on_finalize` only if the current block number is > 1. However, since the block number starts at 0, this will skip the `on_finalize` call at block number 1.
```rust
fn run_to_block(n: u64) {
    while System::block_number() < n {
        if System::block_number() > 1 {
            ExamplePallet::on_finalize(System::block_number());
            System::on_finalize(System::block_number());
        }
        System::set_block_number(System::block_number() + 1);
        System::on_initialize(System::block_number());
        ExamplePallet::on_initialize(System::block_number());
    }
}
```

For example, substituting `Timestamp` (`pallet_timestamp`) for `ExamplePallet` and calling `Timestamp::set()` right after setting the block number will lead to the error `'Timestamp must be updated only once in the block'` in any call with `n >= 2`.

I can open a PR and change the third line in the example to 
```rust
        if System::block_number() > 0 {
```

### Are you willing to help with this request?

Yes!

PR #940, opened 2022-04-09
Update runtime -> testing `run_to_block` example

0xangelo:
This addresses #931


NukeManDan (approved):
Thanks for the detailed issue and PR 😎😁

PR #941, opened 2022-04-10
Update runtime > Frame > skeleton of a pallet code block index.mdx

mikolaichuk:
It looks like "type" keyword is missed here


sacha-l (approved):
🙏🏻 Thanks for catching this.

Issue #558, opened 2021-11-17
Pro-tips for Rust Development 🦀 (for Substrate builders)


NukeManDan:
### _Please add your pro-tips in the comments below!_

There are a few things you learn along the way when building in our ecosystem. A page that is focused on a few key things would help new users accelerate and make their dev day-to-day work more fun 💃🕺

## Key topics:

- IDEs and tooling
  - https://rust-analyzer.github.io/
    - How to get Substrate working (macro issues
- Cargo extensions
  - https://lib.rs/crates/cargo-expand for macro understanding
  - https://lib.rs/crates/cargo-multi for lots of sub-dir actions (like clean && build  on updating toolchains)
  - https://lib.rs/crates/cargo-sweep for nuanced build cleanups to get space back and faster builds
- Build & Test & Deploy Automation
  - CI-CD examples
- ***More topics? Please comment!***

apopiak:
`cargo remote` with a faster build machine

kianenigma:
shamelessly promoting my own little snippet here: https://hello.kianenigma.nl/posts/for-those-who-don-t-want-rust-analyzer-one-regex-to-rul-them-all/, uses 0 cpu clocks and works just as good as RA in msot common cases to find symbols.

wigy-opensource-developer:
Linking substrate binaries need **lots** of RAM with LLVM `lld` command used by cargo. On a 16GB laptop you could easily run out of memory if you also have a browser open on that laptop. Installing and using [`mold`](https://github.com/rui314/mold) I had success doing a full build:

`mold --run cargo build --release`

apopiak:
`cargo tree` for resolving dependency tree issues

NukeManDan:
https://substrate.meta.stackexchange.com/ should surface this issue - anyone wanna take a stab?! :grinning:

PR #966, opened 2022-04-26
Update templateModule front end code example

nodi-andy:
The code for templateModule gives an runtime error (api connection, same issue : #558). The last version:
"https://github.com/substrate-developer-hub/substrate-front-end-template/blob/main/src/TemplateModule.js" works fine.


NukeManDan:
I think this page is not in main-md branch...
we will archive  main now and publish of main-md FYI.

PR #970, opened 2022-04-28
Remove unclear section in Nick tutorial

sacha-l:
The step this PR is removing is confusing: it seems like it's from an older write-up trying to take the shape of this updated version. These aren't types but attributes of the pallet macro, which even if we corrected the terminology I don't think adds much value.


lisa-parity (approved):
Agree that it is a weird step trying to make sense of what was there originally.

lisa-parity:
Are you also making the change in the main-md branch?

Issue #965, opened 2022-04-22
ink workshop source code cannot compile v3


apikic:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### What browsers are you seeing the problem on?

Chrome

### Bug report for the website

It seems that link for source code in https://docs.substrate.io/tutorials/v3/ink-workshop/pt3/#create-the-token-supply are linked to https://github.com/substrate-developer-hub/substrate-contracts-workshop/blob/master/2/assets/2.1-template.rs which cannot be built (as it is described in the steps below in the document). (eg `ink_storage::collections is private module`). 

https://github.com/substrate-developer-hub/substrate-docs/blob/main/v3/tutorials/08-ink-workshop/c-build-erc20-contract/index.mdx is containing the links to the archived repo https://github.com/substrate-developer-hub/substrate-contracts-workshop/

lisa-parity:
Thanks for reporting this bug. I ran into this as well. The original template code doesn't compile until after you update the code with the missing sections (fill in the ACTIONS with the appropriate code). I found that a bit frustrating as a starting point so I created a replacement "template" that would compile.
[template-lib-0-rs.txt](https://github.com/substrate-developer-hub/substrate-docs/files/8585793/template-lib-0-rs.txt)

Now, I just need to decide whether to fix the bug by having a template that doesn't compile (and remove the test and contract UI parts) or by updating the tutorial to point to a more complete "template" that compiles.

And @sacha-l , @NukeManDan, any thoughts? 
Also, where would I put new assets if I want to replace the template (so as not to point to an archived repo, which also confused me)? Maybe here:
https://github.com/substrate-developer-hub/substrate-docs/pull/971

sacha-l:
#972 should fix this :)

PR #971, opened 2022-04-28
Add files via upload

lisa-parity:


sacha-l:
@lisa-parity is this intended to be for `main` ?

NukeManDan:
In think this is intended to patch #965 or other reported ink issues, right?

PR #1001, opened 2022-05-09
Add make to Ubuntu/Debian dependencies

doubledup:
On a fresh Ubuntu 20.04 machine, I saw build failures for the substrate node template. Eventually I found this [StackExchange question](https://substrate.stackexchange.com/questions/1454/failed-to-compile-on-linux/1464#1464), which showed that `make` was not installed. Apparently it doesn't come preinstalled, at least in Ubuntu 20.04.


doubledup:
Added in #1081

PR #1004, opened 2022-05-10
Update obsolete migration label and link

ruibinx:
The label "D1-runtime-migration" is changed to "E1-runtimemigration".


NukeManDan (approved):
Thanks!

PR #1034, opened 2022-05-18
Add default-features = false `frame-try-runtime`

Chralt98:


Chralt98:
Related to https://github.com/zeitgeistpm/zeitgeist/pull/615

NukeManDan:
Hey @sacha-l @Chralt98  should live in the new docs? If you think so please open a new PR targeting the new docs site? Or at least an issue to flag that we should bring this in as new content :pray:

PR #1036, opened 2022-05-19
Update index.mdx

Chralt98:
https://github.com/zeitgeistpm/zeitgeist/pull/615

After having many problems with the integration, I would like to update the code here for others.


NukeManDan (approved):
Thanks for the catch here, I see you use the most up-to-date info from substrate now: https://github.com/paritytech/substrate/blob/master/bin/node-template/runtime/src/lib.rs#L516-L529

PR #1037, opened 2022-05-19
Add `soramitsu/scale-codec-js-library`

0x009922:
## Proposal description

There is a list of SCALE codec implementations. I propose to append [soramitsu/scale-codec-js-library](https://github.com/soramitsu/scale-codec-js-library) into it. This project has a core library and a compiler for it. The whole stack is used in [Iroha 2 JavaScript SDK](https://github.com/hyperledger/iroha-javascript/tree/iroha2/packages/data-model). It is well-tested, but it needs more real-life usage experience to become better. I hope adding it to the implementations list might help with it.

Probably it is **the most performant** and **the most type-safe** JS implementation.

See its [online documentation](https://soramitsu.github.io/scale-codec-js-library/).


appetrosyan (approved):

NukeManDan:
This page will be depricated in the main-md transition (happening soon!)

All libraries like this will be referenced and maintained at https://github.com/substrate-developer-hub/awesome-substrate - please check there for this lib and update if needed!

PR #1040, opened 2022-05-20
fix https in metrics issue

NukeManDan:
https://substrate.stackexchange.com/questions/2738/error-reading-prometheus-post-https-localhost9090-api-v1-query-dial-tcp


vjgaur:
the fix is working, you can close the issue

PR #1089, opened 2022-06-03
kitties tutorial: fix sp_io import

bernardoaraujor:
importing `sp_io::hashing::blake2_128` needs to happend inside `frame_support`.
all the template files were importing outside, which generates
```
error[E0433]: failed to resolve: use of undeclared crate or module `sp_io`
  --> pallets/kitties/src/lib.rs:12:9
   |
12 |     use sp_io::hashing::blake2_128;
   |         ^^^^^ use of undeclared crate or module `sp_io`
```

### DIFF
--- a/static/assets/tutorials/kitties-tutorial/01-basic-setup.rs
+++ b/static/assets/tutorials/kitties-tutorial/01-basic-setup.rs
@@ -7,9 +7,8 @@ pub mod pallet {
     use frame_support::{sp_runtime::traits::{Hash, Zero},
                         dispatch::{DispatchResultWithPostInfo, DispatchResult},
                         traits::{Currency, ExistenceRequirement, Randomness},
-                        pallet_prelude::*};
+                        pallet_prelude::*, sp_io::hashing::blake2_128};
     use frame_system::pallet_prelude::*;
-    use sp_io::hashing::blake2_128;
 
     // TODO Part II: Struct for holding Kitty information.
 
--- a/static/assets/tutorials/kitties-tutorial/02-create-kitties.rs
+++ b/static/assets/tutorials/kitties-tutorial/02-create-kitties.rs
@@ -9,9 +9,9 @@ pub mod pallet {
 	use frame_support::{
 		sp_runtime::traits::Hash,
 		traits::{ Randomness, Currency, tokens::ExistenceRequirement },
-		transactional
+		transactional,
+		sp_io::hashing::blake2_128,
 	};
-	use sp_io::hashing::blake2_128;
 	use scale_info::TypeInfo;
 
 	#[cfg(feature = "std")]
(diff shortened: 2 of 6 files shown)


NukeManDan:
@bernardoaraujor - can this be propagated to main-md? @sacha-l please review :)

sacha-l:
Thanks for looking into this. `sp_io` is not a crate in `frame_support` - I may be missing something but this wouldn't compile as is. 

The fix here is to tell users that they need to go and change the `Cargo.toml` file to specify that `sp_io` will be required in the pallet's `[dependencies]` (in `pallets/pallet/Cargo.toml` for example):

```toml
[dependencies]
codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = [
	"derive",
] }
scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
frame-support = { default-features = false, version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.23"}
frame-system = { default-features = false, version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.23" }
frame-benchmarking = { default-features = false, version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.23", optional = true }
sp-io = { default-features = false, version = "6.0.0", git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.23" }
sp-runtime = { default-features = false, version = "6.0.0", git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.23" }
pallet-randomness-collective-flip = { default-features = false, version = "4.0.0-dev", git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.23" }
pallet-balances = { default-features = false, version = "4.0.0-dev", git = 'https://github.com/paritytech/substrate.git', branch = "polkadot-v0.9.23" }
```

A node template out of the box usually has `sp_io` in `[dev-dependencies]` so all users need to do is move it out of there and into `[dependencies]`. Step 2 of [this section](https://docs.substrate.io/tutorials/v3/kitties/pt1/#write-the-pallet_kitties-scaffold) stipulates this. Perhaps it's not clear enough and we can be more explicit about "why and what" this is about.

bernardoaraujor:
indeed, my PR is going in a wrong direction. I was actually just missing the step 2.

doing these changes does get the code to compile, most likely because `sp_io` [is re-exported](https://github.com/paritytech/substrate/blob/b49d8ee54f4bd6e4b5b48f45d08f75705fad95b4/frame/support/src/lib.rs#L47) by `frame_support`.
regardless, that's not the proper way to do it.

I'll just send another PR clarifying `[dependencies]` vs `[dev-dependencies]`.
sorry for the noise.

bernardoaraujor:
@NukeManDan @sacha-l is kitties not included in `main-md`?
I can't seem to find it.

PR #1097, opened 2022-06-06
rustdocs -> redirect

NukeManDan:
should not be merged until the https://docs.substrate.io/rustdocs/ page is updated to completely removed in favor of only having a redirect. If we keep the page, only need the `/rustdocs/*` redirect IIUC

should also make this onto the `main-md` branch, or target that instead after launch.


NukeManDan:
To do:
- [ ] update Polkadot wiki #1103 to use github pages links
- [ ] scan and update any parity repos to do the same

NukeManDan:
https://github.com/paritytechmarcomms/substrate-website/pull/252/ to replace this... @je-boska why do both repos maintain seporate redirects? Is this explicitly needed because of the subdomian? If so, don't we need _both_?

je-boska:
> [paritytechmarcomms/substrate-website#252](https://github.com/paritytechmarcomms/substrate-website/pull/252) to replace this... @je-boska why do both repos maintain seporate redirects? Is this explicitly needed because of the subdomian? If so, don't we need _both_?

@rmnprkrl Perhaps you could help provide some clarity here.

rmnprkrl:
@NukeManDan we're going to decommission `/rustdocs` page as part of the `main-md` release, we do not need to update the current `main` branch in production, could you re-open this PR against `main-md` (FYI @jonnysmillie )

PR #1007, opened 2022-05-12
add-hrmp-how-to-skeleton

al3mart:
Adding a how-to-guide on how to open hrmp channels.

Content is just a skeleton, needs polishing and relevant changes to follow the style of the rest of documentation.

PR #1099, opened 2022-06-07
Add HRMP guide

sacha-l:
This is a follow up to #1007, in order to include it in `main-md`.

Original author: @al3mart.


sacha-l:
@al3mart : is there anything to add from [this](https://hackmd.io/ECywP_7_SLeTvWCTfQPkqw) ? I haven't looked in much detail, but if separate guide we can do it in a separate PR.

al3mart:
@sacha-l the linked guide covers the steps to open channels with a common good parachain, concretely what is shown is openning channels with Statemine.
The info I provided shows how two community parachains can stablish channels between each other.
So, both documents compliment each other, yet some of the steps to follow are the same.

sacha-l:
Ok - are we good to merge this as is then? cc @al3mart

al3mart (approved):

PR #1108, opened 2022-06-11
/main-docs/ to / - excessive URL levels removed

NukeManDan:
Not sure if this is going to work (cannot do a development build locally at the moment), but IMHO there should be no `/main-docs/ ` level in the URL. We are explicitly on `docs.substrate.io`. Less URL length is better!


jonnysmillie (commented):
Hi Dan, apologies for not reviewing this in good time. I agree with the fundamental preference of not having `/main-docs/` as a category but it would also have some implications for the navigation and site information architecture as it would become inconsistent with the other items that are currently siblings (e.g. Tutorials, Reference, etc). Happy to revisit this and scope it out in future if you also think worthwhile but I would suggest it is part of a larger reworking of the IA of the whole site - where perhaps we separate out pages that are not part of 'main-docs'.

I do definitely agree this is an issue and there are many ways we can approach it. The main complication being keeping a consistent navigation that represents the different content sections of the site appropriately.

1. We rename "Main Docs" to something more appropriate (e.g. "Substrate", "User Guide", "Developer guide", "Setup Guide" and then implement redirects for all `/main-docs/` routes. This would probably be the easiest.
2. We remove "Main Docs" as a folder and have all the items that are currently children of `main-docs` live as siblings of "Quick start", "Tutorials", "Reference". This would also be quite simple but we would require additional design assets for the new navigation items.
3. We rework the architecture of all the top level items so that they exist in some other navigation. This would likely be more complicated and be dependent on other design decisions.

PR #1109, opened 2022-06-11
Remove unused YAML header fields

NukeManDan:
Cleaning up YAML headers.


NukeManDan:
not needed, eventually remove featured image will be in #1114

PR #1111, opened 2022-06-13
fundamentals section review

NukeManDan:
# WIP

don't merge yet

### content/md/en/docs/main-docs/fundamentals/architecture.md

IS THIS CORRECT?  Separation of concerns for outer node <> runtime comms:

- Executor is interface for calls runtime -> outer node
- Runtime APIs are  interface for calls outer node -> runtime

- [ ] Telemetry page missing?
- [ ] high-level or high level?
- [ ] Substrate "outer node" is (almost) completely defined in the `sc-*` crates in the `client` dir. Point this out? Also call outer node client? or perhaps substrate upstream needs a rename. 
- [ ] Light client nodes IMHO doesn't fit here, it's not architecture specifically of a node
- [ ] "A light client or light node is a simplified version of a Substrate node that only provides the runtime and current state." what is current specifically? It doesn't need a very heavy full sync from genesis (I think) What other properties define it? (new page on this and smoldot is of course called for and AFAIK in progress?
- [ ] Light clients can add graphic with end users + various nodes on a network (this is a network architecture diagram, perhaps fits here?) Dan has some ideas! Might be best in the next section though! Maybe this is where it should all live.

### content/md/en/docs/main-docs/fundamentals/networks-and-blockchains.md

- [ ] Network boundaries in libp2p are VERY fuzzy, it's easy to have no "firewall" here. Discuss ProtocolID issues for chains that DOS each other (solo and parachians) as well as the early polkadot overlap with ipfs nodes? What about malicious nodes on your network?
- [ ] "Regardless of the type of chain you want to create, you should plan for deploying to a test network before deploying to a live production network. For example, there are there are currently test networks like Kusama for deploying real-world economics in a canary network and test networks like Rococo for deploying a chain without issuing tokens or having any real-world economic impact. Polkadot is a production relay chain." doesn't fit here. this is design/ build details, not fundamentals.
- [ ] Local development section doesn't fit either: "The nodes that participate in your blockchain define the boundaries of your network. For example, if you are developing a runtime on a local computer, you must install the software required for a Substrate node on the local computer. In most cases, you'll run that local node in development mode by specifying the `--dev` command-line option. In development mode, the node creates a temporary directory to store state while the chain is running. The state in the temporary directory is discarded each time you stop the node. Development mode is primarily for testing and experimentation."
- [ ] --light is deprecated on substrate nodes
- [ ] --pruning is a build detail


lisa-parity:
Hey there, can you focus your efforts on content issues (and not on formatting of bullet lists and tables)? That feedback is far more valuable. Also, can you limit PRs to single files whenever possible?

NukeManDan:
- [ ] define "transaction requests", this is unclear as a term. I think canonically to Substrate you mean "submit extrinsic". If we want to migrate from that term, should likely not exist and all instances of the term would be called transactions instead.
- [ ] "each node stores a copy of the blocks that make up the canonical chain." this is incorrect, not all nodes do (light clients)/
- [ ] "each block and the order in which transactions are processed." - order is not necessarily defined in tx, it is defined in block number/height though (probably should use height or number uniformly in all docs, not both?)
- [ ] "Cryptography for signing and verifying the signatures associated with transactions." - this includes much more use cases for sure. Might be more broad and vague here.
- [ ] "An execution environment for authoring and finalizing blocks." - is ex env = runtime here? finalizing blocks in what way? authoring != finalizing != runtime's job
- [ ] " a complete copy of an existing blockchain repository—a fork—" - I would say source code, more accurate and broad (this reads as git cannonically with those terms repo & fork IMHO)
- [ ] why talk about how to make a new chain in "what is a node" section? seems to impl specific... same with next paragraph, should be new section "why substrate"
- [ ] "a majority of the nodes must agree on all of the state transitions" - that may be true, it is not always. not for likely** needed for finality.
- [ ] "In centralized networks, a central authority can choose between mutually exclusive state transitions by recording the changes to state transition in the order it sees them, and choosing the first of the competing alternatives when a conflict arises." super non-general. This could be one of many ways to do it. Ordering doesn't matter for example in credit card fraud, an arbitrary tranition occurs in reversing that wrt order
- [ ] "The method that a blockchain uses to batch transactions into blocks and to select which node can submit a block to the chain is called the blockchain's consensus model or consensus algorithm. " - no, it's the block authoring model/system.
- [ ] "most consensus models require at least two-thirds of the nodes to agree on state at all time." - no, PoW has 50%. most byzantine fault tolerant models have 2/3 IIRC
- [ ] "Network participants can then deposit the tokens to create a stake of funds that enable them to pay for transactions." stake is canonically not this. Why use that term?
- [ ] "Most blockchains also enable network participants to submit and vote on proposals that affect network operations or the blockchain community." false. Maybe most substrate chains, even that is flaky.
- [ ] "Applications that run on a blockchain—often referred to as decentralized applications or dApps—are typically written as smart contracts." no, they typically are web apps that communicate with a blockchain, maybe contracts, maybe not.
- [ ] " the outcome of programmatically-executed transactions is recorded and can never be changed." kinda? but really what do you mean?
- [ ] "If you prefer to explore code directly, you can start building in the Developer Playground and consult the API reference to get details about the Rust crates you use." - what is dev playground? link?

PR #1129, opened 2022-06-16
Remove incorrect kitty check in mint token htg

hzy1919:


NukeManDan:
This is not useless, explain why it is if you are confident - might be missing something.

hzy1919:
> This is not useless, explain why it is if you are confident - might be missing something.

well, I don't see any code about 'Kitty' in context as well as in https://github.com/substrate-developer-hub/substrate-how-to-guides/blob/main/example-code/template-node/pallets/mint-token/src/lib.rs . So I think it is a fungible-token example not NFT like Kitty . 
I'm not quite sure if I understand correctly .

NukeManDan:
Cc @sacha-l I missed that this was in this specific file, not the tutorial that I suppose it was derived from? @hzy1919 thanks for the spot, although we are moving to archive this version on the site #1132 is about to go live with the main-md branch and this may exist in the new content.

NukeManDan (approved):

hzy1919:
You're welcome! :smiley: @NukeManDan @sacha-l

PR #1142, opened 2022-06-20
Fix typos

hzy1919:


NukeManDan (approved):
fyi main is going to archive, main-md will replace it - please make all PRs to that branch (it will be renamed and the default once live)

hzy1919:
> fyi main is going to archive, main-md will replace it - please make all PRs to that branch (it will be renamed and the default once live)

Ok.

PR #1140, opened 2022-06-17
Update local relay and parachain tutorials

NukeManDan:
- fix media
- update chain specs v0.9.24
- reorder for relay tuto and para tuto separation

PR #1143, opened 2022-06-20
Fix chain specs path

davxy:
Correct chain specs path for both relay chain and parachains


NukeManDan (approved):
#1140  will include these too - fyi main is going to archive, main-md will replace it - please make all PRs to that branch (it will be renamed and the default once live)

PR #1150, opened 2022-06-21
`yarn format` commad

NukeManDan:
covers all markdown files in the contend/md/en/... dir.

Per request, the MD tables are not touched, a latter PR will add ignore comments so we can pass CI with this as a check for formatting


package.json
```diff
@@ -14,7 +14,7 @@
 		"serve": "gatsby serve",
 		"serve:fresh": "yarn clean; yarn build; yarn serve",
 		"clean": "gatsby clean",
-		"format": "prettier --config .prettierrc.js --write \"{!**/*.{js,jsx,json},**/*.{md}}\"",
+		"format": "prettier --config .prettierrc.js --write content/md/en/docs/**/*.md",
```
NukeManDan, line 17: here is what it does

PR #1166, opened 2022-06-23
hide blockchain basics

NukeManDan:
will open an issue to re-introduce this with suggested edits as well. 

#1111  will resolve issues and bring this back


lisa-parity (changes requested):
I'm working with Joe directly. I don't want to hide the topic.

PR #1174, opened 2022-06-23
Fixing bash command because there was a problem with backlashes

dejanzivanov:
`./target/release/node-template key insert --base-path /tmp/node01 \ --chain customSpecRaw.json \ --scheme Ed25519 \ --suri <your-secret-key> \ --password-interactive \ --key-type gran`

was throwing out:

```
error: Found argument ' --scheme' which wasn't expected, or isn't valid in this context

USAGE:
    node-template key insert [OPTIONS] --key-type <KEY_TYPE> --scheme <SCHEME>

For more information try --help
```
So just removing backslashed fixed it.


NukeManDan (approved):
Thanks for the PR! :tada:

PR #1240, opened 2022-07-09
Lg/evm replacement

lisa-parity:
Replacement for the old frontier workshop. While most of it "works" sort of, there are still some problems to be resolved. 
1) There's a mismatch between the create function parameters and the account definition
image
      "Transaction": {
         "nonce": "U256",
         "action": "String",
         "gas_price": "u64",  > maxFeePerGas?
         "gas_limit": "u64", > gasLimit?
         "value": "U256",
         "input": "Vec<u8>",
         "signature": "Signature"
      },

2) I don't think the transfer actually worked. The instructions were pretty vague about what to expect and how to verify that the 221 tokens were actually transferred from alice to bob. Checking accountStorages seemed to always return the same value and I'm not sure if the slot for bob should be 0 or 1 (or something else). I was able to use the evm-slot utility but I don't know if I entered the right parameters to get the expected result. 
Would be great to get some eyes looking at/testing this but my main goal is to provide something so the site doesn't 404 on the missing tutorial.

3) I couldn't figure out how to use the Remix thing, so I couldn't really test that. I took a "best guess" at describing it.


content/config/nav.yaml
```diff
@@ -192,6 +181,19 @@ tutorials:
...
+      url: /tutorials/connect-other-chains/
+      pages:
+        - title: Start a local relay chain
+          url: /tutorials/connect-other-chains/local-relay/
+        - title: Connect a local parachain
+          url: /tutorials/connect-other-chains/local-parachain/
+        - title: Connect to Rococo testnet
+          url: /tutorials/connect-other-chains/rococo-slot/
+        - title: Access EVM accounts
+          url: /tutorials/connect-other-chains/access-evm-accounts/
```
NukeManDan, line 194: how is this related to connection to other chains?

NukeManDan (changes requested):

lisa-parity:
Enables interaction with EVM dapps/accounts. There are references to Moonbeam.
I'm not saying this is a perfect fit. I might add an integrations section later. But for the moment, I think this fits loosely into this category. Or I could remove it from the navigation altogether.

NukeManDan:
Frontier allows you to use EVM smart contracts. It also allows you to have the same cryptography for accounts, and an RPC that matches Ethereum so 3rd party apps can communicate with your node in the same way (interacting with EVM, via smart contracts)

 How is it _not_ a good fit for the contracts section? It has nothing at all to do with bridges, parachains, or any other cross consensus ideas. Just happens that Moonbeam uses this.

shawntabrizi:
I think bikeshedding the location of the content is not that important.

I do have a mental association with Frontier and EVM with the contracts world, but I think the whole tutorial section will probably need a revamp at some point anyway.

I think you should push it through if this is just the one outstanding problem.

shawntabrizi:
At some point, the best thing would be a plan for the tutorial section, including the tutorials that have already been created, and those which are not yet created, and the mental map of where all of them should be.

This could also be influenced by the "tutorial revamp" that I wanted @NukeManDan and @jonnysmillie to drive with the new kitties tutorial.

NukeManDan (approved):
OK, I trust Lisa's vision, even if I cannot quite grasp it yet 🙏

lisa-parity:
the structure will continue to evolve, but for now this update is primarily intended to fix a 404.

PR #1252, opened 2022-07-14
Update tutorial to match current style

lisa-parity:


DrW3RK:
I just noticed this PR after creating this one https://github.com/substrate-developer-hub/substrate-docs/pull/1255

I've suggested a few changes that keep up with the latest substrate node template. I've tested them and they work as expected.

PR #1253, opened 2022-07-14
Update contracts pallet dependencies for the dev hub node template

lisa-parity:


lisa-parity:
looking to sanity check the tutorial...

PR #1266, opened 2022-07-20
WIP: remove main-docs content folder, restructure nav

jonnysmillie:
* Remove `/main-docs/` from urls.
* Remove navigation icons.
* Restructure navigation.
* Redirects for all `/main-docs/` routes.
* Link changes for internal links with`/main-docs/` in the url.


rmnprkrl:
to me this looks great and the IA makes sense!

I would consider to improve the UI to add also the border to the footer, and overflow the nav items that way

image

NukeManDan:
supersedes https://github.com/substrate-developer-hub/substrate-docs/pull/1108

rmnprkrl:
this situation needs to be handled

image

rmnprkrl:
code looks good!

I still think these are a bit of week points in the UI:

1. how the nav layout overflows the content in the middle of nowhere, could be solved by adding the line between sidebar and footer and cut there:
image

2. the submit button imo gets too much space, could be at least smaller (like Docs header button), or more subtle design.
image

lisa-parity:
I think the left nav looks good.
I'd like the Submit an issue and Ask a question links to not be buttons-just text links without the box seems cleaner. I think they might be better served at the bottom of the page with a separator or at the bottom of the right nave so clearly not part of what's on this page.
I'd prefer getting rid of the two colored and two sized buttons for Previous and Next.
Maybe something like this:
------------------------------------------------------------------------------------------------------------------
<        Submit a issue                                                                                                                             Ask a question        >

(Above the standard black footer)

I like the idea of combining the Edit this page and last edited date, but wonder if it is too subtle. Maybe use the Edit icon (with tool tip) instead of the GitHub icon. Or go back to Edit this page, or Edit (link) and date no link?

rmnprkrl:
the overflow should be set all the way down to the border with footer
image

PR #1282, opened 2022-07-22
add custom event handler

rmnprkrl:


rmnprkrl:
will be added to production first in separate PR

PR #1310, opened 2022-07-26
Correct typos in accounts-addresses-keys.md

gflow33:
Added "by" to "This secret seed phrase can be used __ an account owner to recover access to an account if the private key is lost."

Removed the extra 'T' in the first word in "TIn addition to allowing you to derive multiple addresses from the same public key, base-58 encoding has the following benefits"

Removed an extra 'm' in command in "You can look the a chain-specific address for a public key using the subkey inspect commmand and --network option ot by using Subscan." I also think the sentence can be worded differently to convey its meaning.


lisa-parity (approved):

lisa-parity:
Thanks!

PR #1322, opened 2022-07-28
Lg/install code blocks

lisa-parity:
Was missing the language identifier in some places in the macOS, windows, and linux pages.
However, we're reusing the fence block language identifier for the tabbed code blocks (Debian, Arch etc.).
I'm not sure if we can add the Copy functionality for those code blocks as a "custom" solution. Not the worst thing in the world if we can't because they are intended to be *examples* and not necessarily exactly the command that users should copy and paste (it is pretty impossible to know the specific minimum requirements for every flavor and version of Linux out in the world),

### DIFF
--- a/content/md/en/docs/main-docs/install/linux.md
+++ b/content/md/en/docs/main-docs/install/linux.md
@@ -19,7 +19,7 @@ sudo apt install build-essential
 
 At a minimum, you need the following packages before you install Rust:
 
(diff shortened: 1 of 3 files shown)


platonicsocrates:
yeah I think they make sense as examples! Well-spotted!

NukeManDan (approved):

PR #1326, opened 2022-07-28
Add instructions for adding to PATH

platonicsocrates:
Add instructions for adding to PATH + fix typo :)

Missing out the PATH is a very common error we have seen blocking our students from getting set up with the quick start. Once spotted, it's a super simple fix but easy to overlook!

### DIFF
--- a/content/md/en/docs/main-docs/install/troubleshooting.md
+++ b/content/md/en/docs/main-docs/install/troubleshooting.md
@@ -112,7 +112,7 @@ WASM_BUILD_TOOLCHAIN=nightly-<yyyy-MM-dd> cargo build --release
 This command builds the _runtime_ using the specified nightly toolchain.
 The rest of project is compiled using the _default_ toolchain, that is, the latest version of the `stable` toolchain that you have installed.
 
-### Downgrade the nightly toolcahin
+### Downgrade the nightly toolchain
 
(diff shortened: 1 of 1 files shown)


content/md/en/docs/main-docs/install/troubleshooting.md
```diff
@@ -122,3 +122,12 @@ rustup uninstall nightly
...
 ```
+
+### Ensure PATH is set correctly
+
+If after installing Rust the commands don't seem to work, showing errors such as `command not found: rustup`, make sure it your PATH is configured correctly.
+
+Currently, the `rustup` installer installs by default to the bash profile (on mac). If you are using another shell, make sure to add this line to your profile (e.g. `.zshrc`):
+```
+. "$HOME/.cargo/env"
+```
```
NukeManDan, line 133:
```suggestion
```bash
source "$HOME/.cargo/env"
```
```

NukeManDan (changes requested):
I think this is a good addition, do you agree with the change here?

content/md/en/docs/main-docs/install/troubleshooting.md
platonicsocrates: Awesome! Great suggestion - agreed!

lisa-parity (approved):
LGTM

PR #861, opened 2022-03-01
`BoundedVec` PoE update

NukeManDan:
Tested with:
- https://github.com/substrate-developer-hub/substrate-node-template/tree/tutorials/solutions/proof-of-existence
- https://github.com/substrate-developer-hub/substrate-front-end-template/tree/tutorials/solutions/proof-of-existence

~~may update further for https://github.com/multiformats/cid use in FE and in docs https://github.com/multiformats/js-multiformats~~
Tried a while getting https://www.npmjs.com/package/ipfs-only-hash to work (too long)... no luck after a while stuck at errors in the upstream IIUC.

PR #1332, opened 2022-07-29
update from #861 in main-md

NukeManDan:
WIP will also update based on @JoshOrndorff 's workshop at the Academy

### DIFF
--- a/content/md/en/docs/tutorials/work-with-pallets/use-macros-in-a-custom-pallet.md
+++ b/content/md/en/docs/tutorials/work-with-pallets/use-macros-in-a-custom-pallet.md
@@ -116,12 +116,9 @@ Therefore, the first step is to remove some files and content from the files in 
 
    All of the pallets used in a runtime must be set to compile with the `no_std` features.
 
-1. Add the following skeleton code that includes a few minimal [macros for FRAME V2](https://paritytech.github.io/substrate/master/frame_support/attr.pallet.html):
+1. Add the following code that includes a minimal skeleton for a FRAME pallet and its required [macros](https://paritytech.github.io/substrate/master/frame_support/attr.pallet.html):
 
    ```rust
-   //! Substrate Proof-of-Existence Pallet
-   #![cfg_attr(not(feature = "std"), no_std)]
-
    // Re-export pallet items so that they can be accessed from the crate namespace.
    pub use pallet::*;
 
--- a/content/md/en/docs/tutorials/work-with-pallets/use-macros-in-a-custom-pallet.md
+++ b/content/md/en/docs/tutorials/work-with-pallets/use-macros-in-a-custom-pallet.md
@@ -2,6 +2,12 @@
 title: Use macros in a custom pallet
 description: Create a custom pallet for a Substrate runtime using a skeleton of FRAME macros.
 keywords:
+  - proof of existence
+  - PoE
+  - IPFS
+  - hash
+  - root
+  - CID
 ---
 
 This tutorial illustrates how to create a custom pallet for a Substrate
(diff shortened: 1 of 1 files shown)


content/md/en/docs/tutorials/work-with-pallets/use-macros-in-a-custom-pallet.md
```diff
@@ -110,29 +116,46 @@ Therefore, the first step is to remove some files and content from the files in
 
    All of the pallets used in a runtime must be set to compile with the `no_std` features.
 
-1. Add a skeleton set of pallet dependencies and [macros](/reference/frame-macros) that the custom pallet requires by copying the following code:
+1. Add the following skeleton code that includes a few minimal [macros for FRAME V2](https://paritytech.github.io/substrate/master/frame_support/attr.pallet.html):
 
    ```rust
+   //! Substrate Proof-of-Existence Pallet
+   #![cfg_attr(not(feature = "std"), no_std)]
+
```
sacha-l, line 124:
```suggestion
```
This is in step 1 already.

content/md/en/docs/tutorials/work-with-pallets/use-macros-in-a-custom-pallet.md
```diff
@@ -110,29 +116,46 @@ Therefore, the first step is to remove some files and content from the files in
 
    All of the pallets used in a runtime must be set to compile with the `no_std` features.
 
-1. Add a skeleton set of pallet dependencies and [macros](/reference/frame-macros) that the custom pallet requires by copying the following code:
+1. Add the following skeleton code that includes a few minimal [macros for FRAME V2](https://paritytech.github.io/substrate/master/frame_support/attr.pallet.html):
```
sacha-l, line 119:
```suggestion
1. Add the following code that includes a minimal skeleton for a FRAME pallet and its required [macros](https://paritytech.github.io/substrate/master/frame_support/attr.pallet.html):
```
FRAME is FRAME unless we explicitly need to refer to a previous version :)

sacha-l (commented):
I think that step 7 should only contain the macros (i.e. we should revert to the previous layout we had, it makes more sense for a user  who is learning this for the first time to see the bare bones and build ontop of that). Will jump into make a commit to fix this.

content/md/en/docs/tutorials/work-with-pallets/use-macros-in-a-custom-pallet.md
NukeManDan: I see the style Lisa is using is _super extra_ explicit in steps... think we should pull this into it's own step? IMHO no.

sacha-l:
Should be good, just unsure about the  rust indentation - did a `fmt` but maybe double check :) cc @NukeManDan

JoshOrndorff:
So the main change is switching from `Vec<_>` to `BoundedVec<_, _>` right? What do you think about switching to `H256` or `H128` instead? It is a hash that we're storing afterall.

content/md/en/docs/tutorials/work-with-pallets/use-macros-in-a-custom-pallet.md
```diff
@@ -119,41 +119,33 @@ Therefore, the first step is to remove some files and content from the files in
...
-   		/// Event documentation should end with an array that provides descriptive names for event
-   		/// parameters. [something, who]
-   		SomethingHappened(),
-   	}
-   	// TODO: add #[pallet::error] block
-
-   	// TODO: add #[pallet::storage] block
-
-   	// TODO: add #[pallet::call] block
-   }
```
NukeManDan, line 136: @sacha-l this was intentional, not formatting :grimacing:

NukeManDan: specifically, this code ***compiles*** - so we need `Event` defined, though we replace it in a moment

NukeManDan:
> So the main change is switching from `Vec<_>` to `BoundedVec<_, _>` right? What do you think about switching to `H256` or `H128` instead? It is a hash that we're storing afterall.

IIUC, we want to be as generic as possible, so we store an arbitrary vector of `u8`:
```rust
BoundedVec<u8, T::MaxBytesInHash>,
```
So you wanna do something a bit more concrete... is there a `Hash` Trait we have somewhere?

NukeManDan:
#910 should be used as ref. and closes #884

content/md/en/docs/tutorials/work-with-pallets/use-macros-in-a-custom-pallet.md
```diff
@@ -372,8 +359,9 @@ To implement this logic in the proof-of-existence pallet:
    cargo check -p node-template-runtime
    ```
 
-You can refer to the node template [solution](https://github.com/substrate-developer-hub/substrate-node-template/tree/tutorials/solutions/proof-of-existence) if you get stuck.
-You can also check the [commit diff](https://github.com/substrate-developer-hub/substrate-node-template/commit/773d72752f3598e5d405b48c6f716f4153c95070#diff-f191d0eb0afa874a64fb9be550a275f957e220b9076c87d4085c2797ca4e310c) to see the exact changes from the base template.
+> **_Something not working?_**
```
lisa-parity, line 362:
```suggestion
```
Please move these solution links to the end. Including it here implies that you *expect* it not to work. That means it is a bad tutorial. It's find to provide a solution reference, but we've just had them verify that it compiles, so this seems weird.

NukeManDan:
Closing in favor of https://github.com/substrate-developer-hub/substrate-docs/pull/1352#issuecomment-1203593609

PR #1346, opened 2022-08-01
Fix typo in chain-spec.md

gflow33:
I think "them" would be more appropriate than "then" in the context of this sentence... "When distributing chain specs in JSON format, you should distribute **them** in the raw format...."


lisa-parity (approved):
Thanks!

PR #1349, opened 2022-08-02
Update macro doc

sacha-l:
With the deprecated `generate_storage_info` attribute now assumed by default, this updates our macro reference doc to note that an override may be used with caution.


shawntabrizi (approved):

lisa-parity (approved):
LGTM (I'm updating this section so will pull this in first).

PR #1352, opened 2022-08-02
Update Proof of Existence Tutorial

shawntabrizi:
Also updated: https://github.com/substrate-developer-hub/substrate-node-template/blob/tutorials/solutions/proof-of-existence/pallets/template/src/lib.rs

PR #1356, opened 2022-08-03
Fix error in JS documentation code for use macros in a custom pallet tutorial

AlexD10S:
In [this tutorial to make a custom pallet](https://docs.substrate.io/tutorials/work-with-pallets/use-macros-in-a-custom-pallet/#build-a-custom-front-end-component) there is an error with the JS code for the front-end-component: 

`Uncaught TypeError: Cannot read properties of undefined (reading 'toString')`

This is because before add any claim is trying to get the owner and block from an undefined value:

```
setOwner(result.value[0].toString())
setBlock(result.value[1].toNumber())
```

I have added a code that is checking if this value exists before trying to get the owner and the block.


lisa-parity:
Thanks for the fix. I'm moving it to a new front-end tutorial and closing this PR in favor of the rewrite in https://github.com/substrate-developer-hub/substrate-docs/pull/1352

PR #1366, opened 2022-08-04
updated security.txt

soumil1:


static/security.txt
```diff
@@ -1,5 +1,5 @@
 Contact: bugbounty@parity.io
-Preferred-Languages: en
+Preferred-Languages: english
```
NukeManDan, line 2: does this actually matter?

soumil1: Not much, just in case someone is unfamiliar with the abbreviation

NukeManDan: https://en.m.wikipedia.org/wiki/List_of_ISO_639-1_codes

This is very familiar to almost everyone as an industry standard, i would hope that it is rather obvious, and if this file was used in some engine to parse the data, it would include the iso code over the full length version.

NukeManDan:
Thanks for the pr, but I don't think we want this.

static/security.txt
soumil1: Thank You for your insight

PR #1367, opened 2022-08-04
Lg/revise frame macros

lisa-parity:


content/md/en/docs/reference/frame-macros.md
```diff
@@ -1,118 +1,206 @@
 ---
-title: Frame macros
+title: Runtime macros
```
shawntabrizi, line 2: I think keeping the title FRAME is more accurate, since these macros are NOT required for every runtime development. just stuff that uses frame

shawntabrizi: what is your take on why the title would change?

lisa-parity: I was thinking that the context of the discussion was about using the macros in a runtime, but the main motivation was to avoid using proper nouns and acronyms in headings so that headings focus on the topic at hand and not so much on Parity-specific or Substrate-specific terminology.

content/md/en/docs/reference/frame-macros.md
```diff
@@ -228,217 +296,292 @@ impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
...
 
-- [The Rust Programming Language ch 19.5 Macros](https://doc.rust-lang.org/book/ch19-06-macros.html)
+- [The Rust Programming Language](https://doc.rust-lang.org/book/ch19-06-macros.html)
 - [The Little Book of Rust Macros](https://danielkeep.github.io/tlborm/book)
 
-> As of January 2021, FRAME based pallets have upgraded their use of macros. Refer to [this guide](https://paritytech.github.io/substrate/master/frame_support/attr.pallet.html#upgrade-guidelines)
-> to learn about migrating a v1 pallet to v2 and [this resource](https://hackmd.io/@fscJ0GEvRb2rqALLZB1kfA/SkSerkamt) to learn more about FRAME's version
-> changes.
+<!--
+As of January 2021, FRAME based pallets have upgraded their use of macros.
```
shawntabrizi, line 575: delete this?

content/md/en/docs/reference/frame-macros.md
```diff
@@ -1,118 +1,206 @@
...
 
-The purpose of this article is to give a basic overview of Rust macros and explain the Substrate macros
-that runtime engineers most frequently use.
+This section provides an overview of the types of macros available in Rust and highlights how the specific FRAME macros are used in runtime development.
 
 ## Macro basics
 
-Put simply, macros are lines of code that write code. They provide the ability to abstract things out without needing to declare complex data structures explicitly.
-There are four kinds of macro in Rust:
+In computer programming, macros are lines of code that encapsulate a preset sequence of instructions to execute.
```
shawntabrizi, line 14:
```suggestion
In computer programming, macros are code that encapsulate a preset sequence of instructions to execute. 
```

content/md/en/docs/reference/frame-macros.md
```diff
@@ -1,118 +1,206 @@
...
-- in `sp_version`: [function-like](https://paritytech.github.io/substrate/master/sp_version/index.html#macros) macros
+The `frame_support` crate provides many of the most important declarative, derive, attribute-like, and function-like macros used in the runtime.
+A few of the important macros that you should be familiar with from the `frame_support` crate include the following:
 
-See the `dependencies` in the `Cargo.toml` file of the node template runtime to see where these macros are used.
+- `construct_runtime`	used to construct runtime from the list of pallets you have implemented.
+- `decl_error` used to declare an error type for a pallet in the runtime.
+- `decl_event`	used to implement event for a pallet in the runtime.
+- `decl_module`	used to declare a Module data structure and a Call enumeration for a dispatch.
+- `decl_storage` used to declare strongly-typed wrappers around codec-compatible types in storage.
```
shawntabrizi, line 57: we should remove all references to these `decl_*` macros

content/md/en/docs/reference/frame-macros.md
```diff
@@ -1,118 +1,206 @@
...
+### FRAME system macros
+
+The `frame_system` crate uses macros to define primitives that provide access to core data types and shared utilities. 
+These primitives and associated macros form the foundation for many node operations both in the outer node and in the runtime and act as the base layer for other pallets to interact with the Substrate framework.
+
+A few of the important primitives and macros that you should be familiar with from the `frame_system` crate include the following:
+
+- [`sp_core`](https://paritytech.github.io/substrate/master/sp_core/index.html)
+  
+  - `map`	used to initialize a key-value collection from array.
```
shawntabrizi, line 72: i think this is not used anymore, also part of the old macros?

content/md/en/docs/reference/frame-macros.md
```diff
@@ -1,118 +1,206 @@
...
+  
+  - `decl_runtime_apis`	used to declare specified traits as runtime APIs.
+  - `impl_runtime_apis`	used to tag specified trait implementations as runtime APIs.
+  
+  For more information about `sp_api` function-like macros, see [Macros](https://paritytech.github.io/substrate/master/sp_api/index.html#macros).
+
+- [`sp_std`](https://paritytech.github.io/substrate/master/sp_std/index.html)
+  
+  - `if_std`	used to indicate code that should only be run when the `std` feature set is enabled.
+  - `map`	used to initialize a key-value collection from array.
```
shawntabrizi, line 97: duplicate and not used?

content/md/en/docs/reference/frame-macros.md
```diff
@@ -1,118 +1,206 @@
...
-### #[frame_support::pallet]
+### #[pallet]
 
-**When to use**
+The `#[pallet]` macro is required to declare a pallet.
+This [attribute macro](https://paritytech.github.io/substrate/master/frame_support/attr.pallet.html) describes the attributes used to identify the specific items the pallet requires.
+For example, a pallet typically includes a set of types, functions, and trait implementations that are aggregated by the `construct_runtime!` macro to build the runtime.
 
-Required to declare a pallet consisting of a set of types, functions, and trait implementations that will later be aggregated by `construct_runtime` to build to runtime.
+The `#[pallet]` attribute macro defines a pallet as a module:
```
shawntabrizi, line 122: this sentence does not make sense to me

shawntabrizi: macros can only act on code within some structure. the `#[pallet]` macro is an attribute on a `mod pallet`, which is an entry point for all the other `#[pallet::xx]` macros to act on other parts of code.

shawntabrizi: The fact we wrap the code in a `mod pallet` is just an artifact of how we can make attribute macros to work

lisa-parity: Ah, that's much clearer than what I was trying to interpret from the Rust doc.

content/md/en/docs/reference/frame-macros.md
shawntabrizi: yeah delete it, there is only FRAME v2 in our eyes

content/md/en/docs/reference/frame-macros.md
shawntabrizi: that one is okay :)

The FRAME v1 macros are all declarative, whereas the frame v2 are all attribute macros.

Beyond that, just the ones that use `decl_xx` instead of `#[pallet::xx]`

shawntabrizi:
Generally i think you can merge this when you feel happy.

I would say that this is a big more writing into macros than i would probably have done, but i dont see harm.

lisa-parity (commented):
Question about `decl_runtime_apis` and about FRAME v1 - v2 - v3

lisa-parity:
Yeah, it might be overkill, but iI found it useful to dig in a little and it felt like this might be a "glue" topic because specific macros are referenced in a lot of other topics.

PR #1378, opened 2022-08-08
[How-to] Update how to guide to use the define-benchmark syntax instead of add…

lisa-parity:
…-benchmark


lisa-parity:
@ggwpez, if you get a change to take a look here is some updated doc to replace add-benchmark (and other changes).

lisa-parity (commented):
wip

PR #1385, opened 2022-08-09
Update branch that works

lisa-parity:


NukeManDan:
You cannot mix versions in dependencies. Is the intention that all code referencing a version should point to 0.9.26? If so, those should be updated, otherwise users will inevitably run into versioning problems.

PR #1387, opened 2022-08-10
update toolchain version

cuteOlaf:


NukeManDan (approved):
Sure, although this is an _example_ as the "something like" in the line above tries to imply, not an indication of the toolchain you _should_ be using :stuck_out_tongue_winking_eye:

PR #1396, opened 2022-08-13
fix: your-secret-key to your-secret-seed

cuteOlaf:


NukeManDan (changes requested):
key is correct, it could be either.

PR #1398, opened 2022-08-14
Update permissioned-network.md

cuteOlaf:


NukeManDan (changes requested):
features is correct in TOML context

Issue #1429, opened 2022-08-17
Broken links on /build/application-dev/ page


jonnysmillie:
### Broken Link

frame-metadata 
parity-scale-codec 
Storage metadata
the storage RPC
Call metadata
Event metadata
Error metadata
Exrinsic metadata 
AuthorApi
ChainApi
OffchainApi
StateApi
SystemApi

### Context and Origin

Various `https://paritytech.github.io/substrate/master/...`

PR #1451, opened 2022-08-19
Fix broken links found in #1429

sacha-l:
Now that frame-metadata is moved out of Substrate, I've updated the relevant links to point to its `latest` version as published on crates.io.

For the links to `sc_rpc` apis, there was a change in the naming so updated the links accordingly. Please review @lisa-parity for sanity check and that linking to docs.rs is the right thing to do.


lisa-parity (approved):
I'm planning to do a major revision of this topic, but we can go ahead and fix these broken links.
Question though: how can you tell that the API should be Server or Client? I'm wondering because there are a lot of broken links caused by the split out into client/server but I was guessing that most of them would be replaced by links to the ApiClient side of the house (but as I said, I was **guessing**).

PR #1465, opened 2022-08-22
WIP: Skip navigation button

je-boska:
This is an accessibility feature. Users who tab through the page elements, can skip the navigation by clicking this otherwise hidden button as they start tabbing.


je-boska:
@rcos7 Could you implement this as a part of the Substrate plugin instead?

je-boska:
DEPRECATION NOTICE
Due to changes in the site architecture, this feature will be implemented elsewhere, with this PR used for reference only.

PR #1495, opened 2022-08-31
Fix broken URL for application dev

AlexiaChen:
the actual URL is https://docs.substrate.io/build/application-dev/


lisa-parity (commented):
application-development is the new correct name for the file

lisa-parity:
If you found somewhere that the old name was used, please feel free to re-open.

PR #1523, opened 2022-09-20
update glossary

NukeManDan:
See https://github.com/paritytech/substrate/pull/12092#discussion_r971801115

```
- `Dispatchable` is what we call function objects/functors (i.e. data which can be called) in Substrate in general.
- `Call` is the strict name of this data-structure in pallets (an enum).
- `Extrinsic` is a piece of data which can be included in a block and which leads to some action. Typical Substrate chains have extrinsics which contain a `Call` value.
- `Transaction` is an extrinsic which includes a signature so we know which account is authorising it.
```

### DIFF
--- a/content/md/en/docs/reference/glossary.md
+++ b/content/md/en/docs/reference/glossary.md
@@ -117,7 +117,8 @@ The communication overhead for such systems is `O(n²)`, where `n` is the number
 
 ## call
 
-A `Call` is the strict name of a [dispatchable](#dispatch)data-structure in pallets (i.e. an enum).
+In a general context, a call describes the act of invoking a function to be executed.
+In the context of pallets that contain functions to be dispatched to the runtime, `Call` is an enumeration data type that describes the functions that can be dispatched with one variant per pallet. The object that a `Call` represents is a [dispatch](#dispatch) data structure or a dispatchable.
 
 ## collator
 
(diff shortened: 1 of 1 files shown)


content/md/en/docs/reference/glossary.md
```diff
@@ -115,6 +115,10 @@ The loss of a network service due to node failures that exceed the proportion of
 An early approach to byzantine fault tolerance. pBFT systems tolerate byzantine behavior from up to one-third of participants.
 The communication overhead for such systems is `O(n²)`, where `n` is the number of nodes (participants) in the system.
 
+## call
+
+A `Call` is the strict name of a [dispatchable](#dispatch)data-structure in pallets (i.e. an enum).
```
lisa-parity, line 120:
```suggestion
In a general context, a call describes the act of invoking a function to be executed.
In the context of pallets that contain functions to be dispatched to the runtime for execution, `Call` is an enumeration data type that describes the functions that can be dispatched with one variant per pallet.
```

lisa-parity (approved):

content/md/en/docs/reference/glossary.md
NukeManDan: Much cleaner :grin:  - should we try to link explicitly to dispatch though? as a concept these are tightly defined together.
`- `Dispatchable` is what we call function objects/functors (i.e. data which can be called) in Substrate in general.` no _just_ something to be executed IIUC.

NukeManDan: quite open to whatever you think is best, and go ahead and commit to this PR as you see fit and merge when you think it's in a good place :pray:

PR #1544, opened 2022-09-26
fix broken image

evilrobot-01:
Image currently shows as "<Image is missing>", updated path as per working image lower down.


evilrobot-01:
It may be worth cropping the referenced image as per the others on the page, but not sure if that image is used anywhere else.

lisa-parity (approved):

PR #1591, opened 2022-10-15
Update add-a-pallet.md

revoltez:
Nicks pallet uses Event instead of RuntimeEvent in polkadot-v0.9.28.
won't compile if it's RuntimeEvent as it is not defined.
check https://github.com/paritytech/substrate/blob/polkadot-v0.9.28/frame/nicks/src/lib.rs


lisa-parity:
Better to move forward than to move back, so closing this PR in favor of updating the tutorial for v0.9.30.

revoltez:
its confusing for beginners that just copy and paste that's it

PR #1602, opened 2022-10-22
Replace Convert with WeightToFee in the custom inclusion fee code snippet

lisa-parity:


lisa-parity:
I'm not sure if this change is correct because the referenced PR (and others) seem to be about trait WeightToFee  and subtrait WeightToFeePolynomial (I didn't find a reference for a Convert trait).
The only place I've seen Convert was in this release analysis post:
https://forum.parity.io/t/polkadot-release-analysis/1419/2#span-stylefont-size24pxcolor-34b3f1font-weight-600text-decoration-underlineadd-constfeemultiplier-to-the-transaction-payment-palletspanhttpsgithubcomparitytechsubstratepull12222-22

PR #1603, opened 2022-10-24
Change Call, Event, and Origin to RuntimeCall, RuntimeEvent, and RuntimeOrigin

lisa-parity:


sacha-l (commented):
Left review notes here only to explain the follow-up review commit I made. Hope it makes sense.

PR #1625, opened 2022-11-16
Add documentation of event unit testing

elv-nate:
I noticed that it was quite hard to find information on how to test for event emission, and the best resources were just stackoverflow posts that were fairly hard to find. As such, I added this documentation and also added an example of testing for an event to the node template.

Related PR on the node template: paritytech/substrate#12722.


lisa-parity (approved):

lisa-parity:
@elv-nate, thanks for suggesting this addition!

PR #1628, opened 2022-11-18
Comment section out

sacha-l:
Caught some text that was probably meant to be notes. Looking at this article, we could probably list some useful traits from https://paritytech.github.io/substrate/master/frame_support/traits/index.html#traits

Happy to add some commits to improve this article, though I could use a little direction. Is the purpose of this article to be a sort of landing page that branches out to docs like macros, pallet coupling, different existing pallets or is it meant to provide examples of composing a runtime with combining existing pallets + some custom pallet?

We can also merge this for now and improve in a follow-up PR, let me know.


lisa-parity (approved):

lisa-parity:
I think the goal of the article was probably to cover the why and how to use the scaffolding for a custom pallet, but I haven't looked deeply into it (mostly a pull over from the old site with maybe some light editing—I haven't been through all of the topics that require improvement ;-D).

PR #1724, opened 2023-01-10
Update add-trusted-nodes.md

gogosama-dev:
Amended keystore outputs in order to resolve identical key hashes confusion.


lisa-parity (approved):
Thanks!

PR #1772, opened 2023-01-31
Update simulate-parachains.md

brunopgalvao:
Update for PBA!

### DIFF
--- a/content/md/en/docs/test/simulate-parachains.md
+++ b/content/md/en/docs/test/simulate-parachains.md
@@ -48,7 +48,7 @@ To prepare a working folder with the binaries for the test network:
 2. Clone the Polkadot repository by running a command similar to the following:
    
    ```bash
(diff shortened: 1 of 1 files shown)


pepoviola (approved):
Thanks @brunopgalvao!

brunopgalvao:
@lisa-parity this is bumping the version to `0.9.37` and has minor updates to the zombienet config.

NukeManDan:
@lisa-parity good time to update all parachain content, as the present docs are quite out of date. Should only be minor things to tweak throughout.

lisa-parity (approved):

NukeManDan:
Sorry if it was not clear, i think this info update does not work with the other parachain content using 0.9.32. as the versions are too old. Is that the case @brunopgalvao ?

If so this pr should be reverted imho, because its a footgun for users to fight with version mismatches and usually they don't find that is the problem without substantial headache first.

PR #1789, opened 2023-02-08
[FIX] Update simulate-parachains.md

gogosama-dev:
Issue fixed:
- Removed duplicate of the entire article
- Fixed broken links
- Line 179: Fixed relative directory path (binaries are NOT in substrate-parachain-template folder)
- Line 244, 250: Fixed deprecated statement (up-to-date version is snake_case instead of camelCase and has no 'S' at the end)
- Line 268: Fixed typo
- Line 285: Fixed erroneous statement (it's called "xcmPallet" in Polkadot.js portal, not "polkadotXcm", at least in Rocococ testnet)

Question for further improvement:
* Explain or fix indentation in config.toml
* What part of config.toml does [[hrmp_channel]] belong? Shall it be in [relaychain] section? 
* (SUGGESTION) It would be cool to see complete final versions of the files changed in the tutorials, as an attachment or something - similar to what they usually do as general practice in eLearning courses on programming.

### DIFF
--- a/content/md/en/docs/test/simulate-parachains.md
+++ b/content/md/en/docs/test/simulate-parachains.md
@@ -22,312 +22,6 @@ This tutorial illustrates how to set up a basic test network with the following 
 - Four validators
 - Two parachains
 - One collator per parachain
-- One message passing channel that enables the parachains to exchange messages
-
-## Prepare a working folder with the binaries
-
-The `zombienet` command-line interface relies on a configuration file to specify the characteristics of the test network, including the name and location of the binaries, Docker image, or Kubernetes deployment to use.
-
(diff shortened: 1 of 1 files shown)


lisa-parity:
Hi @gogosama-dev, I haven't had a chance to test this yet and I'm not seeing the duplicated text in my main branch so I don't want to blindly approve. 

- I'm not sure if the TOML file requires sections/subsections to be indented. It might just be a visual clue, but as I said, I haven't tested this yet.
- I think hrmp_channel is its own section.
- Unfortunately there are both the "xcmPallet" and the "polkadotXcm" and it depends where you are which one you use--one in the context of the relay chain and the other in the context of a parachain. I'd have to test to see which direction we're going to verify whether it should be `xcmPallet` or `polkadotXcm`.

lisa-parity (approved):

PR #1791, opened 2023-02-08
Fix the URL for the Ruby SCALE codec implementation

wyhaines:
The Ruby library that is linked to in the docs has been deprecated and archived in favor of a newer implementation.

This change just changes the reference to point to the currently supported version for Ruby.


sacha-l (approved):
Thanks for updating this ! 👍🏻  🙏🏻

PR #1794, opened 2023-02-09
Add some links to some SS58 implementations.

wyhaines:
We have a very complete reference of implementations for Base58 on our page about Base58, which is very useful for anyone coming at the Substrate ecosystem from another language, or for anyone who wants to see how things work at a low level, but we are lacking the same set of resources for our SS58 checksumming / address format implementation.

This list is probably incomplete, but it has the 4 most-complete implementations that I have found so far outside of the Substrate and Polkadot-js core implementations, and I think it could be useful, just like the Base58 list, for anyone looking at our ecosystem from outside of it, particularly if they are not Rust programmers, as our docs and our code for our Rust SS58 implementation are pretty dense.


lisa-parity (approved):
thanks

PR #1805, opened 2023-02-10
Fix docker link.

wyhaines:
It was inserted as a relative link instead of an absolute link.


lisa-parity:
Hey @wyhaines, have you successfully set up an archive node? I tried to follow the instructions and couldn't get it working. I think there's something missing / assumed most likely about the PostgreSQL DB configuration.

lisa-parity (approved):

PR #1861, opened 2023-02-24
Add security headers for dev env

miabarbir:


jonnysmillie (approved):
Looks good to me Mia, though not sure why this branch is out of sync with main. Don't think it matters :)

PR #1890, opened 2023-03-09
add instruction for the default() constructor in Use maps for storing values tutorial

bencgreenberg:
The Use maps for storing values tutorial builds from the previous tutorial and asks the reader to continue adding more functionality on top of what they did previously. However, it is missing a step to update the `default` constructor that the reader built in the previous tutorial, so the tutorial as it is now does not build and the tests cannot run.

This PR adds the instruction to update the `default` constructor with the `Mapping` like the rest of the tutorial.


lisa-parity (approved):

HCastano:
@bencgreenberg the [sample solution for the `Mapping` tutorial](https://github.com/substrate-developer-hub/substrate-docs/blob/main/static/assets/tutorials/smart-contracts/incrementer-mapping.rs) doesn't have a `Default` constructor. I think I removed it on purpose.

Can you double check if we really need to add this? Or maybe we can remove the use of the default constructor in the other parts of the tutorial

PR #1906, opened 2023-03-14
Restructuring draft

har00ga:
Made some changes to the structure that seemed to make sense to me. It would also be cool to break down each of these into related submenus as well, especially with the bigger sections like build and tutorials.

The 'troubleshooting your code' page can probably be broken up into 2 sections; the first half going under tutorials and the second half (error handling) could be it's own section, perhaps under test or build. 'Randomness' can probably also be moved to fundamentals.

The fundamental section could also be split into 2 submenus: blockchain-generic fundamentals, and those specific to Substrate itself, as there's quite a bit of information that depending on a user's original knowledge may or may not be necessary.

Feel like moving or maybe rearranging the tutorials may be good as well, but not sure how I'd like to see that executed as of now.

Any thoughts / comments / concerns?

### DIFF
--- a/content/config/nav.yaml
+++ b/content/config/nav.yaml
@@ -59,24 +59,6 @@ fundamentals:
       url: /fundamentals/light-clients-in-substrate-connect/
     - title: Cross-consensus messaging
       url: /fundamentals/xcm-communication/
-
-install:
-  title: Install
-  url: /install/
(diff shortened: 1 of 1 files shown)


har00ga:
@lisa-parity any thoughts on these adjustments?

PR #1919, opened 2023-03-27
Lg/update nav interim

lisa-parity:


lisa-parity:
Close in favor of https://github.com/substrate-developer-hub/substrate-docs/pull/1925

PR #1956, opened 2023-04-13
Remove instances of insecure randomness

### LINKED ISSUE #1908: Remove all mention & use of insecure randomness pallet
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Content request

https://github.com/paritytech/substrate/pull/13560 has removed the pallet, and shortly the devhub templates will be updated to remove it as well. We _do not_ want people to use the pallet at all, so no examples or mention of it would be best to get in place asap.

Kown references to it:
- [ ] https://docs.substrate.io/build/randomness/
- [ ] https://docs.substrate.io/reference/how-to-guides/pallet-design/incorporate-randomness/

### Are you willing to help with this request?

No.

sacha-l:
Closes #1908.

Did a pass on articles, guides and tutorials to remove all mentions of the insecure randomness pallet as we want to discourage it's use. The only place it's still used is in the Collectibles tutorial which may need to be revised, but imo for creating DNA as a learning piece, it's OK to use the insecure randomness pallet.


content/md/en/docs/build/randomness.md
```diff
@@ -15,10 +15,10 @@ In traditional computer applications, what people call random numbers are actual
 Pseudo-randomness depends on a sufficiently random seed—provided by the user or by an external source—and an algorithm that manipulates the seed to generate a series of seemingly-random numbers. 
 The result is a pseudo-random number that is difficult to predict because of the algorithm used to generate it, but deterministic because the same seed input always produces the same series of output.
 
-However, applications that run on the blockchain are more tightly constrained because all authorities in the network must agree on any on-chain value, including any randomness data that is injected.
-Because of this constraint, you can't use real randomness directly in blockchain applications.
+However, applications that run on the blockchain are more tightly constrained. 
+Pseudo-randomness introduces a security vulnerability for situations where being able to predict the outcome of a random function can lead to gaming the system.
```
Daanvdplas, line 19: I think the part that you removed here is important. So I'd add your new bit to it: "For this reason, Pseudo-randomness introduces a ...."

sacha-l: I'm not sure I understand the statement/ logic saying: "there's a constraint, which is that nodes must agree on any on-chain value and therefore randomness can't be used on chain". It's not because of this constraint that you can't use randomness. Randomness in of itself is the constraint too, where you could argue that true randomness does not exist. 

So I'm not sure what information the lines I'm suggesting to remove are providing. Could you provide bullet points of you're interpretation?

sacha-l, line 19:
```suggestion
However, applications that run on the blockchain are more tightly constrained because all authorities in the network must agree on any on-chain value, including any randomness data that is injected.
Because of this constraint, you can't use real randomness directly in blockchain applications.
For this reason, using randomness on-chain securely must utilize techniques such as VRFs otherwise it becomes possible to predict the outcome of a random function which could lead to gaming the system.
```

sacha-l: That makes sense. Adding an update. Thanks for your comments @Daanvdplas 🙏🏻

Issue #1954, opened 2023-04-11
Broken link


tadeohepperle:
### Broken Link

https://docs.substrate.io/tests/

https://docs.substrate.io/iinstall/

### Expected Location

https://docs.substrate.io/install/
https://docs.substrate.io/test/

### Context and Origin

https://docs.substrate.io/quick-start/

in Install and Learn

PR #1992, opened 2023-06-10
Fixed broken links in Update index.md - #1954

Rashmirreddy:


kianenigma:
What we seem to need is a tool that also checks internal links. I think the github action we added recently can do that, but we explicitly configure it not to.

Rashmirreddy:
Merging these changes. Will also check how to go about the internal link checker.

PR #2038, opened 2023-08-10
Remove the redundant trailing M at Update what-can-you-build.md

behkod:
Doing all this to get rid of a single 'M'! To fix or not to fix?! Symptoms of OCD?!


shawntabrizi:
looks like this was already repaired on master

PR #2046, opened 2023-08-25
replace T::BlockNumber with BlockNumberFor<T>

GitNiko:
It will cause error because associated type `BlockNumber` not found for `T`


ghe0 (approved):
The solution is clean. It also compiles and works (test via UI).

shawntabrizi:
looks like this was already fixed in master

PR #2065, opened 2023-09-20
fix(storage): updated way how to create storage

TomaszWaszczyk:
Looks like documented way how to create storage is not up to date. I got:

image

The PR fixes warnings.

PS. Formatting of MD file is made by VSCode plugin.


shawntabrizi:
this has been fixed in master

PR #2082, opened 2023-10-28
fix: dependency git url inside `authorize-specific-nodes.md`

BeroBurny:
On the Tutorial -> `Build a blockchain` -> `Authorize specific nodes`

There is a wrong URL in the example for using a dependency https://docs.substrate.io/tutorials/build-a-blockchain/authorize-specific-nodes/#add-node-authorization-dependencies


shawntabrizi:
looks like this has been fixed in master

Issue #2091, opened 2023-11-28
Broken git repository link to clone polkadot chain template for a local relay chain


poisonhackerfinger:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Stack Exchange first?

- [X] This is not a support question.

### Bug report for compiling, code snippets, templates, etc.

the command git clone --branch release-v1.0.0 https://github.com/paritytech/polkadot-sdk.git does not work. 
the link in the docs https://github.com/paritytech/polkadot/releases points to a different repository. 
my logical conclusion is the command should be git clone --branch v1.0.0 https://github.com/paritytech/polkadot.git which works.

### Steps to reproduce the problem

git clone --branch release-v1.0.0 https://github.com/paritytech/polkadot-sdk.git

poisonhackerfinger:
Of course!

I found it in this tutorial https://docs.substrate.io/tutorials/build-a-parachain/prepare-a-local-relay-chain/#build-the-relay-chain-node

abujulaibib:
Thanks for reporting, I will create a PR to correct it.👍🏼😊

PR #2097, opened 2023-12-10
Update prepare-a-local-relay-chain.md

abujulaibib:


abujulaibib:
This PR aims to address issue #2091 thereby updating the git clone section to be consistent with the Polkadot version v1.0.0 used throughout the tutorial. However, this can be changed in the future if newer releases are adopted across all sections in the tutorial.

PR #2116, opened 2024-01-31
Improved english?

thiolliere:
Is it better english?


thiolliere:
is it me who is wrong? there is other instance of such grammar like: `the event is not be reverted.`

kianenigma (approved):
It sure is.

PR #2141, opened 2024-03-10
Update acquire-a-testnet-slot.md

mordamax:
Suggest by default to use web version of faucet (as more easier and faster) keeping matrix as an alternative


kianenigma (approved):

kianenigma:
TY

PR #2152, opened 2024-05-14
remove careers.substrate links

carinaluise:


rmnprkrl (approved):
This PR removes https://careers.substrate.io (will be decommissioned) from the website IA.

PR #2162, opened 2024-06-26
Add max storage value size info

apopiak:
Adds a note on suggested rule of thumb for max storage value size.


apopiak:
Seems like this is outdated as parityDB is the default. Is there a limit for parityDB?

